PR #25, opened 2021-04-19
Client should not handle adding required subjects to mentor

kycarr:


kycarr:
@beatthat this is already changed in your branch right?

beatthat:
> @beatthat this is already changed in your branch right?


> @beatthat this is already changed in your branch right?

yeah you can close this

PR #26, opened 2021-04-20
Video Recorder stuff

kycarr:
- fix: home screen wasn't actually waiting for training to complete

WIP:
- add types and api call for uploadVideo and VideoUploadState
- record screen calls uploadVideo and waits for upload+transcribe

### DIFF
--- a/cypress/cypress/support/functions.ts
+++ b/cypress/cypress/support/functions.ts
@@ -1,6 +1,10 @@
 import { Mentor, _Ref } from "./types";
 import { login as loginDefault } from "../fixtures/login";
 import { mentorDefault } from "../fixtures/mentor";
+import { TrainStatus, VideoStatus, UserAccessToken } from "./types";
+
+const TRAIN_STATUS_URL = `/classifier/train/status`;
+const UPLOAD_STATUS_URL = `/videos/upload/status`;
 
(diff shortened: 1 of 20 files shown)


cypress/cypress/support/functions.ts
```diff
@@ -1,6 +1,10 @@
 import { Mentor, _Ref } from "./types";
 import { login as loginDefault } from "../fixtures/login";
 import { mentorDefault } from "../fixtures/mentor";
+import { TrainStatus, VideoStatus, UserAccessToken } from "./types";
+
+const TRAIN_STATUS_URL = `/classifier/train/status`;
```
beatthat, line 6: these should come from config (config can set them, and let these literals be defaults)

beatthat (changes requested):
looks good...smattering of questions and change requests. let me know if questions

cypress/cypress/support/functions.ts
kycarr: @beatthat this is in cypress. The value is a default, the cypress functions take params.statusUrl or use the constant otherwise.

beatthat: oh right! my bad

beatthat (approved):

PR #32, opened 2021-04-26
New branch for slicing videos

kycarr:


client/src/api.ts
```diff
@@ -608,17 +608,20 @@ export async function fetchTrainingStatus(
...
-  data.append("body", JSON.stringify({ mentorId, videoId }));
-  data.append("file", video);
+  data.append(
+    "body",
+    JSON.stringify({ mentor: mentorId, question: questionId })
+  );
+  data.append("video", video);
   const request = axios.create({
-    baseURL: VIDEO_ENTRYPOINT,
+    baseURL: UPLOAD_ENTRYPOINT,
```
beatthat, line 621: I think 10secs might be too low for timeout on a video upload? Maybe 30?

client/src/api.ts
```diff
@@ -634,6 +637,41 @@ export async function fetchUploadVideoStatus(
   return result.data.data;
 }
 
+export async function trimVideo(
```
beatthat, line 640: let's just pass `trimStart` and `trimEnd` as upload params

client/src/pages/record.tsx
```diff
@@ -219,36 +226,47 @@ function RecordPage(props: {
...
+    if (!curAnswer) {
+      return;
+    }
+    setCurAnswer({ ...curAnswer, recordedAt: "" });
+  }
+
   useInterval(
     (isCancelled) => {
-      fetchUploadVideoStatus(statusUrl)
+      fetchUploadVideoStatus(uploadStatusUrl)
```
beatthat, line 256: for the isCancelled below, we only have cancelled as true if this page is unmounted, so don't set any state in that case, just return

client/src/pages/record.tsx
```diff
@@ -271,6 +289,44 @@ function RecordPage(props: {
     isUploading ? 1000 : null
   );
 
+  useInterval(
+    (isCancelled) => {
+      fetchTrimVideoStatus(trimStatusUrl)
+        .then((videoStatus) => {
+          if (isCancelled()) {
+            setIsTrimming(false);
```
beatthat, line 297: don't set state on isCancelled (for unmount), just return

client/src/types.ts
```diff
@@ -181,20 +181,25 @@ export interface VideoStatus {
...
-  UPLOAD_SUCCESS = "UPLOAD_SUCCESS",
-  TRANSCRIBE_STARTED = "TRANSCRIBE_STARTED",
-  TRANSCRIBE_PENDING = "TRANSCRIBE_PENDING",
-  TRANSCRIBE_FAILURE = "TRANSCRIBE_FAILURE",
-  TRANSCRIBE_SUCCESS = "TRANSCRIBE_SUCCESS",
+  FAILURE = "FAILURE",
+  SUCCESS = "SUCCESS",
+  PENDING = "PENDING",
+  STARTED = "STARTED",
+  // NONE = "NONE",
```
beatthat, line 188: delete all the commented lines

cypress/cypress/integration/record.spec.ts
```diff
@@ -557,11 +587,85 @@ describe("Record", () => {
...
   });
 
-  it("shows video recorder if mentor type is VIDEO", () => {
+  it("shows video recorder if mentor type is VIDEO and no video", () => {
     cySetup(cy);
     cyMockDefault(cy, { mentor: mentor2 });
     cy.visit("/record");
     cy.get("[data-cy=video-recorder]").should("exist");
+    // video recorder showing
+    cy.get("[data-cy=video-recorder]").should("have.css", "visibility", "visible");
```
beatthat, line 596: does `should("be.visible")` not work in these cases? If we can use it, let's prefer it over css checks

beatthat (changes requested):
Looks great! change requests are for:

- let's not have a dedicated server-side trim function (can we live w show the crop tools until upload completes?)

- do we need to do the css visibility checks in cypress? if we can use `should("be.visible")` maybe less fragile?

PR #36, opened 2021-05-20
Video recording and cleanup useEffects

kycarr:


client/src/components/nav-bar.tsx
```diff
@@ -28,7 +28,7 @@ import MenuIcon from "@material-ui/icons/Menu";
 
 import { CLIENT_ENDPOINT } from "api";
 import Context from "context";
-import withLocation from "wrap-with-location";
+import withLocation from "hooks/wrap-with-location";
```
beatthat, line 31: is `wrap-with-location` a hook?

beatthat (changes requested):
looks good. my main comment is a tiny one: `wrap-with-location` doesn't seem to be a hook? Unless i'm missing something, let's put only actual hooks (or hook-specific helper code) under `/hooks`

beatthat (approved):

PR #40, opened 2021-06-14
4.1.x - user sees the 'scope' of their virtual mentor to encourage making a better mentor

supersweet-dev:
Shows mentor scope card in question review. Stages encourage users to improve their mentors.


beatthat (approved):
looks good!

PR #41, opened 2021-06-16
4.1.X - Global prettier + Global Licenses

StapleMink:


beatthat (commented):
Did something change about handling of line length?

It looks like now a lot of long lines that used to get broken up just run long?

PR #43, opened 2021-06-18
4.1.x - mentor can track multiple answers simultaneously with option to cancel

aaronshiel:
A completed job is dismissed by either:
- Visiting its respective page or
- Pressing the dismiss x button


client/src/components/record/uploading-widget.tsx
```diff
@@ -0,0 +1,135 @@
...
+    (upload) =>
+      upload.uploadStatus !== UploadStatus.DONE &&
+      upload.uploadStatus !== UploadStatus.UPLOAD_FAILED
+  );
+  const height = 250;
+  const width = 350;
+
+  //the IDx of an answer corresponds to its position within the answers array
+  function retrieveAnswerIDx(id: string) {
+    let i = 0;
```
beatthat: why is this `i` scoped outside the for loop? Shoudn't be this?

```typescript
for (const i = 0; i < answers?.length; i++) {
```

beatthat (changes requested):
few change requests inline. the most important one is the one about not using a form to post the cancel json...do we need that for some reason?

client/src/api.ts
```diff
@@ -625,14 +627,31 @@ export async function uploadVideo(
...
+  mentorId: string,
+  question: Question,
+  taskId: string
+): Promise<CancelJob> {
+  const data = new FormData();
+  data.append(
+    "body",
+    JSON.stringify({ mentor: mentorId, question: question._id, task: taskId })
+  );
+  const result = await uploadRequest.post("/answer/cancel", data);
```
beatthat: I think this should just be:

```typescript
const result = await uploadRequest.post(
  "/answer/cancel", 
  { mentor: mentorId, question: question._id, task: taskId }
);
```

...or if not, why not?

beatthat (approved):

client/src/api.ts
kycarr: it is

PR #45, opened 2021-06-21
Trim sends start and end in seconds instead of as a percentage

kycarr:


beatthat (approved):

beatthat (changes requested):
needs a merge with main and conflicts resolved, then I think this is good to go

PR #50, opened 2021-06-25
makefile fixed

benmai22:
Currently, there's "make format". We should have a new rule "make pretty".

Basically, "make format" will do pretty and license.


beatthat:
hey, don't close pull requests that are in progress. If you want to make clear that it isn't ready for merge, then just change it to `draft`

PR #52, opened 2021-06-25
fixed makefile

benmai22:
Currently, there's "make format". We should have a new rule "make pretty".

Basically, "make format" will do pretty and license.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,8 @@ node_modules/license-check-and-add:
 node_modules/prettier:
 	npm ci
 
-.PHONY: format
-format: node_modules/prettier
+.PHONY: pretty
+pretty: node_modules/prettier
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -37,9 +37,13 @@ LICENSE_HEADER:
...
 	exit 1
 
+.PHONY: format
+format: LICENSE LICENSE_HEADER
+	npm run license:fix && npm run format
+	
 .PHONY: license
 license: LICENSE LICENSE_HEADER
-	npm run license:fix
+	npm run license:fix && npm run format
```
kycarr, line 46: License shouldn't run format as well right?

kycarr (changes requested):

kycarr (approved):

PR #53, opened 2021-06-25
Added fetchThumbnail query, updates thumbnail on upload

supersweet-dev:
needs to pass accessToken from Home component to MyMentorCard component, not sure if this is ok?


client/src/components/my-mentor-card.tsx
```diff
@@ -83,8 +84,8 @@ StageProgress.propTypes = {
 };
 const useStyles = makeStyles(() => ({
   avatar: {
-    width: "100%",
-    height: "100%",
+    width: 240,
```
beatthat, line 87: does this work well across different device resolutions? I'm not a css/layout guy but is there some better way to do this, e.g. `em` or `rem`?

client/src/components/my-mentor-card.tsx
```diff
@@ -225,9 +228,15 @@ export default function MyMentorCard(props: {
                 data-cy="upload-file"
                 type="file"
                 accept="image/*"
-                onChange={(e) =>
-                  uploadThumbnail(props.mentorId, e!.target!.files![0])
-                }
+                onChange={(e) => {
+                  uploadThumbnail(props.mentorId, e!.target!.files![0]).then(
```
beatthat, line 232: no `not-null` assertions ever

beatthat (changes requested):
access-token stuff is weird (it should be a react Context thing), but we can leave it for now because Ben Mai is working on switching everything auth to `https-only` cookies.

I did have one q inline about whether there isn't a better way to size the image than hard code (which presumably works well for only specific target resolutions?) I can be convinced on that if you're sure this is the best way to do it.

Then just fix the non-null assertion

beatthat (changes requested):
looks good to me, the thing it needs now is merge main/resolve any conflicts, then should be good to go.

Also, when you want this reviewed again, make sure to hit the REFRESH button to signal ready for review

beatthat (changes requested):
Looks good, but you are now the first candidate on the new "separate hooks are generally required" rule

So, can we get all the data/hook code for thumbnails broken out into a new custom hook, e.g.

```typescript
[thumbnail, saveThumbnail] = useThumbnail()
```
?

beatthat (approved):
looks great

PR #54, opened 2021-06-28
4.1.x - Centered Headers

supersweet-dev:
Headers are larger and center-aligned. Some pages were re-titled, review answers was moved up to top of sub-category and renamed My Mentor.
Do we like how this looks?


beatthat:
did you want this reviewed?

beatthat:
@supersweet-dev  let me know if this is still active and wants review. If not, we should close/delete

beatthat (approved):

PR #57, opened 2021-06-29
Tips for idle video slide

aaronshiel:


beatthat:
@aaronshiel  is this still active? Aiming to merge eventually or should be close + delete branch?

aaronshiel:
> @aaronshiel is this still active? Aiming to merge eventually or should be close + delete branch?

In Queue. This was just the beginning of the "Mentor sees a setup video of recommended recording setup" user story, but I was put on a different story shortly after.

client/src/components/setup/idle-tips-slide.tsx
```diff
@@ -0,0 +1,51 @@
...
+      title="Recording an idle video."
+      content={
+        <div>
+          <Typography variant="h6" className={classes.text}>
+            In this video, you will learn how to record an idle video and why it
+            is important.
+          </Typography>
+          <ReactPlayer
+            data-cy="video-player"
+            url={"https://youtu.be/xSu1BhuFt8A"}
```
beatthat, line 37: The video URL should be a config setting provided by the server. It can still have this default (defined in the graphql code, but it will be overridable)

See [here](https://github.com/mentorpal/mentor-graphql/blob/9cbdb273aabd5cc61056a4b3fb89c1d5881470fa/node/src/gql/query/config.ts#L13) and follow through to the actual SettingsModel to see how you can do the above

client/src/hooks/graphql/use-with-record-state.tsx
```diff
@@ -342,7 +342,7 @@ export function useWithRecordState(
     upload(mentor._id, answer.answer.question, answer.recordedVideo, trim);
   }
 
-  function cancelUploadVideo(task: UploadTask) {
+  function cancelUploadVideo(task: UploadTask | undefined) {
```
beatthat, line 345: shouldn't this be 

```typescript
function cancelUploadVideo(task?: UploadTask) {
```

?

...curious also what would be the case where task param would be null?

beatthat (changes requested):
looks good. A few comments and change requests inline.

There's also an issue where snapshot images can be flaky and now we're got a blank one in there. I'm wondering if we can fix this or if the snapshot tools are just too unreliable (in which case we may be better to lose the snapshots)

cypress/cypress/integration/setup.spec.ts
```diff
@@ -524,14 +539,13 @@ describe("Setup", () => {
...
   it("shows required subject, repeat after me, questions slide", () => {
-    cySetup(cy);
     cyMockDefault(cy, {
       ...baseMock,
       mentor: [setup6, setup6, setup8, setup8],
       subject: repeatAfterMe,
       gqlQueries: [mockGQL("updateAnswer", true, true)],
     });
-    cy.visit("/setup?i=6");
+    cy.visit("/setup?i=7");
```
beatthat, line 548: OK. Let's actually fix this with a helper function at top of file

```typescript
enum SetupScreen { 
  Idle = 1, 
 //... 
}

function cyVisitSetupScreen(cy, screen: SetupScreen) {
  cy.visit(`/setup?i={number(screen)}`);
}
```

beatthat (changes requested):
looks good. biggest change ask is let's make sure config access are all via redux and whatever `useConfig` hook now

Other small changes requested inline

aaronshiel:
alpha tested on:
(mentor-admin) 4.2.0-alpha.26
(mentor-graphql) 4.2.0-alpha.3

client/src/hooks/graphql/use-with-setup.tsx
```diff
@@ -96,6 +99,7 @@ export function useWithSetup(
     startTask: startTraining,
     clearError: clearTrainingError,
   } = useWithTraining();
+  const config = useWithConfig();
```
beatthat, line 102: I think we probably want `useWithSetup` to bail out immediately if config is not loaded? If we can't do that, we need to change how we handle the case of "config not loaded" below

beatthat (changes requested):
looking good...let's touch base about how to handle `config not loaded` case in hooks that depend on config tho

beatthat (dismissed):

beatthat (dismissed):

beatthat (approved):
looks great

PR #59, opened 2021-06-30
silhouette guide visible while trimming

aaronshiel:


client/src/custom.d.ts
```diff
@@ -0,0 +1,15 @@
+/*
+This software is Copyright ©️ 2020 The University of Southern California. All Rights Reserved. 
+Permission to use, copy, modify, and distribute this software and its documentation for educational, research and non-profit purposes, without fee, and without a written agreement is hereby granted, provided that the above copyright notice and subject to the full license file found in the root of this software deliverable. Permission to make commercial use of this software may be obtained by contacting:  USC Stevens Center for Innovation University of Southern California 1150 S. Olive Street, Suite 2300, Los Angeles, CA 90115, USA Email: accounting@stevens.usc.edu
+
+The full terms of this copyright and license should always be found in the root directory of this software deliverable as "license.txt" and if these terms are not found with this software, please contact the USC Stevens Center for the full license.
+*/
+/*
```
beatthat, line 7: double license again

cypress/cypress/integration/record.spec.ts
```diff
@@ -2261,6 +2261,61 @@ describe("Record", () => {
     cy.get("[data-cy=trim-video]").should("not.exist");
   });
 
+  it("guide silhouette should be visible while trimming a video", () => {
+    cySetup(cy);
+    cyMockUpload(cy);
```
beatthat, line 2266: I would argue having a distinct function `cyMockUpload` separate from `cyMockDefault` and then 33 instances of the following:

```typescript
cyMockUpload(cy);
cyMockDefault(cy);
```

...is a violation of the DRY principal. Any reason not to make mocking the `upload` endpoint a part of the default mock and then getting rid of the distinct `cyMockUpload` (or at least don't export it)?

Probably we should do the same with `cyMockCancelUpload`

cypress/cypress/integration/record.spec.ts
```diff
@@ -2261,6 +2261,61 @@ describe("Record", () => {
...
+              },
+            ],
+          ],
+          true
+        ),
+      ],
+    });
+    cy.intercept("**/videos/mentors/*/*.mp4", { fixture: "video.mp4" });
+    cy.visit("/record");
+    cy.fixture("video.mp4").then((fileContent) => {
```
beatthat, line 2305: ...same comment about DRY violations. Fix globally for the project as follows:

```typescript
// in support functions...

export function cyAttachUpload(
  cy, 
  /* can also be parametrized to override default upload */
): Promise<void> {
  return cy.fixture("video.mp4").then((fileContent) => {
      cy.get('input[type="file"]').attachFile({
        fileContent: fileContent.toString(),
        fileName: "video.mp4",
        mimeType: "video/mp4",
      });
    });
}
```

...I made the above return a `Promise` because I suspect it's also a race condition to not wait for that async attach behavior to complete before continuing?

...so all usages should be more like

```typescript
cyAttachUpload(cy).then(() => {
  // all remaining test code here
})

**Review by beatthat — CHANGES_REQUESTED** [review]

File: client/src/components/record/video-player.tsx [hunk 4]
```diff
@@ -4,7 +4,11 @@ Permission to use, copy, modify, and distribute this software and its documentat
 
 The full terms of this copyright and license should always be found in the root directory of this software deliverable as "license.txt" and if these terms are not found with this software, please contact the USC Stevens Center for the full license.
 */
-
+/*
```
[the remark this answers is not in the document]
aaronshiel: 33259fb

client/src/custom.d.ts
aaronshiel: 33259fb

cypress/cypress/integration/record.spec.ts
aaronshiel: 33259fb

cypress/cypress/integration/record.spec.ts
aaronshiel: 33259fb

beatthat (approved):
looks great

PR #60, opened 2021-06-30
dropdown should now expand when adding a question

aaronshiel:


cypress/cypress/integration/home.spec.ts
```diff
@@ -165,6 +165,17 @@ describe("Review answers page", () => {
     cy.get("[data-cy=stage-toast]").contains("You have 5 total questions.");
   });
 
+  it("dropdown should expand when question is added", () => {
+    cySetup(cy);
+    clint.answers[4].status = Status.COMPLETE;
```
beatthat, line 170: You almost always want to treat data as immutable unless it's lifecycle is completely in the scope of your function.

So here, breaking the isolation of this test and  mutating some `clint` data that is being used by all the other tests in the spec. It's probably also mutating it for tests in other specs, but not 100% sure on that.

The right way to do this would be something like

```typescript
const testClint = {
  ...clint,
  answers: clint.answers.map((a, i) => {
    return (i !== 4) 
      ? a
      : {
          ...a,
          status: Status.COMPLETE
      }
  })
}
```

...I see similar problems--mutating a `clint` object that lives in a shared module--in other tests in this spec file. We should fix them all before merge

beatthat (changes requested):

cypress/cypress/integration/home.spec.ts
aaronshiel: e6323c1

beatthat (approved):
looks great!

PR #61, opened 2021-06-30
[obsolete - delete after my mentor merged ]4.1.x - rec review page contains mentor card  placeholder

supersweet-dev:


client/src/components/my-mentor-card.tsx
```diff
@@ -327,9 +341,178 @@ export default function MyMentorCard(props: {
   );
 }
 
+export function ProgressCard(props: {
```
beatthat, line 344: this file is unmanageably large. Maybe better like

```
components
  my-mentor
    index.tsx. # just imports/exports
    my-mentor-card.tsx
    progress-card.tsx
```

client/src/components/my-mentor-card.tsx
```diff
@@ -327,9 +341,178 @@ export default function MyMentorCard(props: {
...
+}
+
 MyMentorCard.propTypes = {
   mentorId: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
   value: PropTypes.number.isRequired,
   thumbnail: PropTypes.string.isRequired,
 };
+
+ProgressCard.propTypes = {
```
beatthat, line 513: I forget. Do propTypes do anything for us that a typescript interface does not? I guess the proptypes are checked at runtime and not just pre-build is the point?

client/src/pages/record.tsx
```diff
@@ -116,6 +121,12 @@ function RecordPage(props: {
       navigate("/");
     }
   }
+  function onEnd() {
```
beatthat, line 124: want to get away from fetch calls inside components. Looks like there's already a `useRecordState` hook. Can this be moved into there, e.g. `recordState.refetchMentor`?

cypress/cypress/integration/record.spec.ts
```diff
@@ -19,6 +19,7 @@ import {
   completeSubjectQuestion,
   updateMentorAnswer,
 } from "../support/helpers";
+import mentor from "../fixtures/mentor/clint_home";
```
beatthat, line 22: can you double check whether absolute imports work here? Prefer them if they do

beatthat (changes requested):
looks good. Bunch of comments inline.

Biggest ask is can we get a screenshot visual-regression test for `my-mentor` (and any new screen)?

The reason is that as we go along it helps alot in PR reviews to be able to see in the commits what the screen looks like (and if it has changed how)

cypress/cypress/integration/record.spec.ts
supersweet-dev: I think we could probably discard this pr since I redid the bash button in it's own PR and Aaron's suggestion page is going to take the place of this feature, but I can address the best practices changes in the bash-button PR?

beatthat: ok, sounds good. Once we merge your other PR we can close/delete this one

PR #63, opened 2021-07-06
4.2.x - Adds bash button to guide mentor progress.

supersweet-dev:
Button asks user to upload a thumbnail (not implemented), record idle video, and add more subjects.
Working on more features.


client/src/components/bash-button.tsx
```diff
@@ -0,0 +1,149 @@
...
+    },
+  };
+  const idle = mentor?.answers.find(
+    (a) => a.question.name === UtteranceName.IDLE
+  );
+  const categories: {
+    subjectName: string;
+    subject: string;
+    categoryName: string;
+    category: string | undefined;
```
beatthat, line 35: generally better to use the empty string instead of undefined to represent unset and then just test falsiness. Is that not also what you're doing here w `categoryName` anyway?

client/src/components/bash-button.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { Answer, MentorType, Status, UtteranceName } from "types";
+
+export default function BashButton(props: {
+  accessToken: string;
+  setThumbnail: (file: File) => void;
+}): JSX.Element {
+  const { mentor } = useWithReviewAnswerState(props.accessToken, {
+    subject: undefined,
+  });
+  const bash = {
```
beatthat, line 21: can we change the terminology (and all associated names) from `bash` to `recommendedAction`?

client/src/components/bash-button.tsx
```diff
@@ -0,0 +1,149 @@
...
+    text: "",
+    reason: "",
+    action: () => {
+      undefined;
```
beatthat, line 25: does type check or lint require this `undefined` ? If not, lose it

client/src/components/bash-button.tsx
```diff
@@ -0,0 +1,149 @@
...
+            .includes(a.question._id)
+        ),
+      });
+    });
+  });
+  const firstIncomplete = categories.find((c) =>
+    c.answers.find((a) => a.status === Status.INCOMPLETE)
+  );
+
+  switch (true) {
```
beatthat, line 71: Instead of `switch` would be cleaner and easier to understand if you break this out into a function. 

```typescript
interface RecommendedAction {
  text: string;
  reason: string;
  action: () => void
}
function recommend(): RecommendedAction {
  if (!mentor.thumbnail) {
    return {
      text: "Add a thumbnail",
      // ...
    }
  }
}
```

client/src/components/bash-button.tsx
```diff
@@ -0,0 +1,149 @@
...
+      bash.action = () => {
+        navigate(
+          `/record?back=${encodeURI(`/?subject=${undefined}`)}&videoId=${
+            idle?.question._id
+          }`
+        );
+      };
+      break;
+    case Boolean(firstIncomplete):
+      bash.text = "Answer " + firstIncomplete?.categoryName + " Questions";
```
beatthat, line 92: use backtick strings instead of + concat

client/src/components/bash-button.tsx
```diff
@@ -0,0 +1,149 @@
...
+      bash.reason =
+        "You have unanswered questions in the " +
+        firstIncomplete?.subjectName +
+        " subject";
+      bash.action = () =>
+        navigate(
+          `/record?back=${encodeURI(
```
beatthat, line 99: use `URLSearchParams` to construct/encode query strings.

Maybe easier if we had a global helper function , e.g.

```typescript
function urlBuild(base: string, params: Record<string, string>) {
  const query = URLSearchParams()
  Object.keys(params).forEach(n => query.append(n, params[n])
  return `${base}?{query.toString()}`
}
```

client/src/components/bash-button.tsx
```diff
@@ -0,0 +1,149 @@
...
+        {bash.text}
+      </Button>
+      <Typography variant="body1" color="textSecondary" data-cy="bash-reason">
+        {bash.reason}
+      </Typography>
+    </div>
+  );
+}
+
+BashButton.propTypes = {
```
beatthat, line 147: probably we don't need the propTypes?

client/src/components/home/index.tsx
```diff
@@ -87,16 +85,10 @@ function HomePage(props: {
...
-    if (setupStatus?.isSetupComplete === false) {
-      navigate(`/setup`);
-    }
-  }, [setupStatus]);
-  if (!mentor || !setupStatus) {
+  if (!mentor) {
     return (
       <div>
-        <NavBar title="Mentor Studio" mentorId={mentor?._id} />
+        <NavBar title="Mentor Studio" mentorId={undefined} />
```
beatthat, line 91: can this just be empty string when unset?

client/src/hooks/graphql/use-with-setup.tsx
```diff
@@ -2,6 +2,11 @@
 This software is Copyright ©️ 2020 The University of Southern California. All Rights Reserved. 
 Permission to use, copy, modify, and distribute this software and its documentation for educational, research and non-profit purposes, without fee, and without a written agreement is hereby granted, provided that the above copyright notice and subject to the full license file found in the root of this software deliverable. Permission to make commercial use of this software may be obtained by contacting:  USC Stevens Center for Innovation University of Southern California 1150 S. Olive Street, Suite 2300, Los Angeles, CA 90115, USA Email: accounting@stevens.usc.edu
 
+The full terms of this copyright and license should always be found in the root directory of this software deliverable as "license.txt" and if these terms are not found with this software, please contact the USC Stevens Center for the full license.
+*/
+/*
```
beatthat, line 7: double license. Need to check what is causing these

client/src/pages/record.tsx
```diff
@@ -116,6 +119,11 @@ function RecordPage(props: {
       navigate("/");
     }
   }
+  function onEnd() {
```
beatthat, line 122: let's move `refetchMentor` to be managed by the `useWithRecordState` hook

cypress/cypress/integration/mentor-card.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+This software is Copyright ©️ 2020 The University of Southern California. All Rights Reserved. 
+Permission to use, copy, modify, and distribute this software and its documentation for educational, research and non-profit purposes, without fee, and without a written agreement is hereby granted, provided that the above copyright notice and subject to the full license file found in the root of this software deliverable. Permission to make commercial use of this software may be obtained by contacting:  USC Stevens Center for Innovation University of Southern California 1150 S. Olive Street, Suite 2300, Los Angeles, CA 90115, USA Email: accounting@stevens.usc.edu
+
+The full terms of this copyright and license should always be found in the root directory of this software deliverable as "license.txt" and if these terms are not found with this software, please contact the USC Stevens Center for the full license.
+*/
+import { cySetup, cyMockDefault } from "../support/functions";
+import clint from "../fixtures/mentor/clint_home";
+import { Status, QuestionType, UtteranceName } from "../support/types";
+
+describe("Mentor Card on My Mentor Page", () => {
```
beatthat, line 11: I think cleaner to have just one root describe and then you can still have nested describes, e.g.

```typescript
describe("My Mentor Page", () => {
  describe("Mentor Card", () => {})
  describe("Recommended Action", () => {})
})

File: cypress/cypress/integration/mentor-card.spec.ts [hunk 12]
```diff
@@ -0,0 +1,168 @@
...
+  it("does not show toast on incomplete level", () => {
+    cySetup(cy);
+    cyMockDefault(cy, { mentor: clint });
+    cy.visit("/");
+    cy.get("[data-cy=stage-toast]").should("not.exist");
+  });
+
+  it("shows mentor scope toast on stage floor", () => {
+    cySetup(cy);
+    const testClint = {
```
beatthat, line 68: rule of thumb: always inline a variable if you can. The reason to NOT inline would be if other things need to access the variable by name; I think here they don't so...

```typescript
cyMockDefault(cy, { 
  mentor: {
    ...clint,
    // ...
  }
});

File: cypress/cypress/integration/mentor-card.spec.ts [hunk 13]
```diff
@@ -0,0 +1,168 @@
...
+    cy.visit("/");
+    cy.get("[data-cy=stage-toast]").contains(
+      "Your mentor has reached the Scripted stage!"
+    );
+    cy.get("[data-cy=stage-toast]").contains("You have 5 total questions.");
+  });
+
+  it("does not show progress bar at maxed level", () => {
+    cySetup(cy);
+    const testClint = {
```
beatthat, line 89: inline

cypress/cypress/integration/mentor-card.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+  });
+});
+describe("Bash Button Guides User To Improve Mentor", () => {
+  const bashClint = {
+    ...clint,
+  };
+
+  it("Asks user without thumbnail to upload one", () => {
+    cySetup(cy);
+    bashClint.thumbnail = "";
```
beatthat, line 122: never edit anything scoped outside of a test function inside a test function. That is a recipe for race-condition errors and broken isolation

```typescript
cyMockDefault(cy, { 
  mentor: {
    ...bashClint,
    thumbnail: ""
  }
});

File: cypress/cypress/integration/mentor-card.spec.ts [hunk 15]
```diff
@@ -0,0 +1,168 @@
...
+    cyMockDefault(cy, { mentor: bashClint });
+    cy.visit("/");
+    cy.get("[data-cy=bash-thumbnail]").should("exist");
+    cy.get("[data-cy=bash-upload]").should("be.hidden");
+  });
+
+  it("Asks user with thumbnail and without idle video to upload one", () => {
+    cySetup(cy);
+    bashClint.thumbnail = "url";
```
beatthat, line 131: inline and never mutate shared objects

cypress/cypress/integration/mentor-card.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    bashClint.answers.forEach((a) => {
```
beatthat, line 132: once inlined, this will be easier as a map and always should be map anyway

cypress/cypress/integration/mentor-card.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    cyMockDefault(cy, { mentor: bashClint });
+    cy.visit("/");
+    cy.get("[data-cy=bash-button]").contains("Record an Idle Video");
+    cy.get("[data-cy=bash-button]").trigger("mouseover").click();
+    cy.url().should("include", "videoId=idletest");
+  });
+
+  it("Asks user with thumbnail and idle video to answer incomplete questions", () => {
+    cySetup(cy);
+    bashClint.answers.forEach((a) => {
```
beatthat, line 147: inline and no mutations

cypress/cypress/integration/mentor-card.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    cyMockDefault(cy, { mentor: bashClint });
+    cy.visit("/");
+    cy.get("[data-cy=bash-button]").contains("Answer Category2 Questions");
+    cy.get("[data-cy=bash-button]").trigger("mouseover").click();
+    cy.url().should("include", "/record");
+  });
+
+  it("Asks user with thumbnail, idle video, and complete questions to add a subject", () => {
+    cySetup(cy);
+    bashClint.answers.forEach((a) => (a.status = Status.COMPLETE));
```
beatthat, line 161: inline and no mutations (use `map` not `forEach`)

beatthat (changes requested):
looking good. Change requests inline

client/src/components/bash-button.tsx
supersweet-dev: I think this is about the only thing I'm missing. Should we use this function everywhere we're generating urls?

client/src/pages/record.tsx
supersweet-dev: Talked to Aaron and it doesn't look like we need to refetch the mentor live anymore. Removed this instead.

beatthat (approved):
looks great!

PR #64, opened 2021-07-06
4.2.x? - Cleanup setup:

kycarr:
- remove next and back button on slides and replace with arrows
- remove done and save button on slides
- changing slides saves automatically


client/src/pages/profile.tsx
```diff
@@ -2,19 +2,24 @@
 This software is Copyright ©️ 2020 The University of Southern California. All Rights Reserved. 
 Permission to use, copy, modify, and distribute this software and its documentation for educational, research and non-profit purposes, without fee, and without a written agreement is hereby granted, provided that the above copyright notice and subject to the full license file found in the root of this software deliverable. Permission to make commercial use of this software may be obtained by contacting:  USC Stevens Center for Innovation University of Southern California 1150 S. Olive Street, Suite 2300, Los Angeles, CA 90115, USA Email: accounting@stevens.usc.edu
 
+The full terms of this copyright and license should always be found in the root directory of this software deliverable as "license.txt" and if these terms are not found with this software, please contact the USC Stevens Center for the full license.
+*/
+/*
```
beatthat, line 7: double license bug :/

cypress/cypress/integration/setup.spec.ts
```diff
@@ -131,27 +203,25 @@ describe("Setup", () => {
...
-    cy.get("[data-cy=back-btn]").should("be.disabled");
-    cy.get("[data-cy=done-btn]").should("be.disabled");
-    cy.get("[data-cy=radio-0]").should(
-      "have.css",
-      "color",
-      "rgb(27, 106, 156)"
-    );
+    cy.get("[data-cy=radio]")
+      .eq(0)
+      .should("have.css", "color", "rgb(0, 128, 0)");
```
beatthat, line 208: generally want to shy away from testing explicit colors. Is ok if we have a small number of instances

cypress/cypress/integration/setup.spec.ts
```diff
@@ -167,14 +237,14 @@ describe("Setup", () => {
...
         cy.get("input").should("have.value", "");
       });
-      cy.get("[data-cy=save-btn]").should("be.disabled");
     });
-    cy.get("[data-cy=radio-1]").should("have.css", "color", "rgb(255, 0, 0)");
+    cy.get("[data-cy=radio]")
+      .eq(1)
+      .should("have.css", "color", "rgb(255, 0, 0)");
     // fill out first name and save
+    cy.wait(500);
```
beatthat, line 245: makes me nervous that we have `cy.wait`...is here to address test flake? Is there no other way?

beatthat (changes requested):
Overall looks really good. My biggest two questions/requests:

1. cy.wait()

Places where we're using `cy.wait()` is there any other way? I worry about flake when I see it...

2. snapshots/visual regression testing

I think cypress snapshots are still a valuable tool. As a baseline, for a big "UI CHANGE" PR like this, it would help a lot when reviewing to see some snapshots in the file diff, so I can understand what the changes look like in the UI. Also, I can see in many places there are tests for things like css color. Can we use a combination of component snapshots and screen snapshots to capture the important UI states (e.g. this 'XYZ radio button is red when ABC')?

cypress/cypress/integration/setup.spec.ts
kycarr: Yeah, the issue is the slide plugin we're using has an animation that, during which, the buttons below lose focus so cypress can't click them

cypress/cypress/integration/setup.spec.ts
```diff
@@ -35,69 +35,145 @@ Cypress.on("uncaught:exception", (err, runnable) => {
...
       cyMockDefault(cy, baseMock);
       cy.visit("/setup");
-      cy.get("[data-cy=slide]").contains("Welcome to MentorPal!");
+      cy.get("[data-cy=slide-title]").should(
+        "have.text",
+        "Welcome to MentorPal!"
+      );
       cy.get("[data-cy=next-btn]").trigger("mouseover").click();
-      cy.get("[data-cy=slide]").contains("Tell us a little about yourself.");
+      cy.wait(500);
```
beatthat, line 53: ack...what's causing us to need all these `cy.wait` ?

beatthat (changes requested):
the big issue is what to do about all the `cy.wait`...I want to find a way to NOT use cy.wait

Looking here: 

https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Animations

...cypress says they try to wait for animating elements already, but this doesn't work for us?

If we absolutely have to, I would prefer to have a some type of env setting in the client itself that disables animations for cypress testing (at least the problem ones). We would need to make sure this also works in the dockerized e2e tests (I guess env var that is respected at build time in docker)

...but before we do any of the above, are we really sure there's no way to get it to just work in regular cypress w no `cy.wait`?

beatthat (changes requested):
looks great! Let's just take out the snapshots that are just an empty slide for now (or remove all the new snapshots if it's non-deterministic which ones do or don't get populated)

beatthat (approved):
looks great

PR #65, opened 2021-07-07
4.2.x - Recommend follow up question UI

aaronshiel:


client/src/pages/record.tsx
```diff
@@ -164,150 +189,183 @@ function RecordPage(props: {
...
+            <Typography className={classes.title}>Question:</Typography>
+            <FormControl className={classes.inputField} variant="outlined">
+              <OutlinedInput
+                data-cy="question-input"
+                multiline
+                value={curAnswer?.editedAnswer.question?.question}
+                disabled={
+                  curAnswer?.editedAnswer.question?.mentor !== mentor._id
+                }
+                onChange={(e) => {
```
beatthat, line 232: seems like a lot of logic inside of a component's event handler, and then (maybe?) is repeated below w some minor variation? Can this be broken out into a named function (private to record.tsx still)?

cypress/cypress/integration/record.spec.ts
```diff
@@ -559,6 +560,186 @@ describe("Record", () => {
...
+      });
+      cy.get("[data-cy=status]").contains("Skip");
+      cy.get("[data-cy=back-btn]").should("not.be.disabled");
+      cy.get("[data-cy=next-btn]").should("not.exist");
+      cy.get("[data-cy=done-btn]").should("exist");
+      cy.get("[data-cy=done-btn]").trigger("mouseover").click();
+      cy.get("[data-cy=stage-card]").should("exist");
+    });
+
+    it.only("End-of-category page should display unordered list of follow up questions", () => {
```
beatthat, line 606: make sure no `.only` gets pushed to github or especially merged.

beatthat (changes requested):
looks good. There are a bunch of change requests scattered in inline comments. 

One important note is to be sure to double check before/after you push commits that there are no `.only` markers left in cypress. 

Another spanning change request is that we use google's JS style:

https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined

...and by that style book, variable names it should be `mentorId` NOT `mentorID` So that needs to be updated in all `xxxID` var names

client/src/hooks/graphql/use-with-data.tsx
```diff
@@ -87,12 +87,12 @@ export function useWithData<T>(fetch: () => Promise<T>): UseData<T> {
...
   }
 
-  function saveData(update: UpdateFunc<T>) {
+  async function saveData(update: UpdateFunc<T>): Promise<void> {
     if (state.isLoading || state.isSaving || !editedData || !update) {
       return;
     }
     dispatch({ type: LoadingActionType.SAVING, payload: true });
-    update
+    await update
```
beatthat, line 95: when you use `await` you don't need `then` (compiler does that for you). So if you're doing the async version of this would be more like

```typescript
const updated = await update.action(editedData);
if(state.isLoading) {
  //...
```

Let's change the name from `update.callback` to `update.action` if what we're doing is passing an arbitrary action that performs some kind of update.

client/src/hooks/graphql/use-with-record-state.tsx
```diff
@@ -130,7 +130,7 @@ export function useWithRecordState(
       !isMentorLoading
     ) {
       setRecordPageState(RecordPageState.RECORDING_ANSWERS);
-      if (answerIdx < mentor.answers.length) nextAnswer();
+      if (answerIdx < answers.length) nextAnswer();
```
beatthat, line 133: our js/ts style is always use curly braces to explicitly block code, so

```typescript
if (answerIdx < answers.length) {
  nextAnswer();
}
```

client/src/pages/record.tsx
```diff
@@ -423,7 +424,7 @@ function RecordPage(props: {
               disabled={recordPageState === RecordPageState.RELOADING_MENTOR}
               onClick={() => {
                 setRecordPageState(RecordPageState.RELOADING_MENTOR);
-                saveSubject(recordState.reloadMentorData);
+                saveSubject().then(() => recordState.reloadMentorData());
```
beatthat, line 427: break this onClick handler out into a function with a name like `handleSaveSubject` (or whatever is appropriate)

As a rule of thumb, we only inline component event handlers if the code is a single statement

beatthat (changes requested):
looks great. just a few minor tweak change requests in line

client/src/hooks/graphql/use-with-data.tsx
```diff
@@ -21,7 +21,7 @@ const initialState: LoadingState = {
 
 interface UpdateFunc<T> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  callback: (data: T) => Promise<any>;
+  action: (data: T) => Promise<any>;
```
beatthat, line 24: why `Promise<any>` instead of `Promise<void>`...something is weird if callers are expecting return objects w this pattern?

client/src/hooks/graphql/use-with-record-state.tsx
```diff
@@ -115,6 +123,38 @@ export function useWithRecordState(
     }
   }, [mentorError]);
 
+  useEffect(() => {
+    if (!mentor) return;
+    if (
+      recordPageState === RecordPageState.RELOADING_MENTOR &&
+      !isMentorLoading
+    ) {
+      setRecordPageState(RecordPageState.RECORDING_ANSWERS);
```
beatthat, line 132: this is the kind of thing that can potentially cause test flake. Calling `setRecordPageState` here could trigger a render where `recordPageState=RECORDING_ANSWERS` but whatever `nextAnswer` does has not occurred yet. Would that be valid or would it cause errors?

Probably react will batch this update before calling render (as long as they're synchronous), but it's a thing to watch out for

cypress/cypress/integration/home.spec.ts
```diff
@@ -110,6 +110,74 @@ describe("My Mentor Page", () => {
     cy.get("[data-cy=recommended-action-button]").should("exist");
   });
 
+  it("shows placeholder when no thumbnail", () => {
+    cySetup(cy);
+    const testClint = {
```
beatthat, line 115: inline all these `testClint` variables into the mock call. As a general rule, if something can be inlined (because nothing else in scope accesses it), then inline it

cypress/cypress/integration/record.spec.ts
```diff
@@ -4,7 +4,17 @@ Permission to use, copy, modify, and distribute this software and its documentat
 
 The full terms of this copyright and license should always be found in the root directory of this software deliverable as "license.txt" and if these terms are not found with this software, please contact the USC Stevens Center for the full license.
 */
-import { cyMockDefault, mockGQL, cyAttachUpload } from "../support/functions";
+/*
```
beatthat, line 7: remove double license

cypress/cypress/integration/record.spec.ts
```diff
@@ -578,6 +588,141 @@ describe("Record", () => {
       cy.get("[data-cy=done-btn]").should("exist");
       cy.get("[data-cy=done-btn]").trigger("mouseover").click();
     });
+
+    it("End-of-category page should display unordered list of follow up questions", () => {
```
beatthat, line 592: this is super minor, but good to get used to the grammar of BDD testing

DESCRIBE:  recording page
IT: displays a list of followup questions at end of category

...to me is the english phrasing that would make sense

cypress/cypress/integration/record.spec.ts
```diff
@@ -578,6 +588,141 @@ describe("Record", () => {
...
+      );
+      cy.get("[data-cy=follow-up-question-2]").should("exist");
+      cy.get("[data-cy=follow-up-question-2]").should("be.visible");
+      cy.get("[data-cy=follow-up-question-2]").should(
+        "have.text",
+        "What does an Intern do?"
+      );
+    });
+
+    it("If no follow up questions are generated, final button should not traverse to follow up questions page", () => {
```
beatthat, line 696: "does not traverse to followups page if there are no followups"

beatthat (changes requested):
looks good. A few change requets and one question in line. Mostly minor stuff

client/src/api.ts
```diff
@@ -586,6 +587,17 @@ export async function trainMentor(mentorId: string): Promise<AsyncJob> {
   return result.data.data;
 }
 
+export async function fetchFollowUpQuestions(
+  categoryId: string,
+  accessToken: string
+): Promise<FollowUpQuestion[]> {
+  const headers = { Authorization: `bearer ${accessToken}` };
+  const result = await axios.get(
```
beatthat, line 595: we need to change this to POST if that's what it is on the classifer-api side

beatthat (changes requested):
make the classifier/followup call POST to match classifier-api

beatthat (approved):
let's merge once we see it all work together (at least locally)

PR #66, opened 2021-07-07
video display correctly updates during trimming

aaronshiel:


client/src/components/record/video-player.tsx
```diff
@@ -71,19 +72,23 @@ function VideoPlayer(props: {
...
-    if (duration && state.played >= trim[1] / 100) {
+    if (duration && state.played >= trim[1] / 100 && !trimInProgress) {
       reactPlayerRef.current.seekTo(duration[0]);
     }
   }
 
   function onUpdateTrim(value: number | number[]): void {
     if (!trimInProgress) setTrimInProgress(true);
-    if (!Array.isArray(value) || !reactPlayerRef?.current) {
+    if (
```
beatthat, line 82: is confusing what `value` is here since it can be essentially a scalar number or what I guess is really a tuple of two numbers (start, end)? What is the case for it ever being a scalar?

beatthat (commented):
see inline, wondering if there is a way to clarify the code a little by tweaking that param name `value` and it's allowed types

client/src/components/record/video-player.tsx
```diff
@@ -77,19 +77,17 @@ function VideoPlayer(props: {
     }
   }
 
-  function onUpdateTrim(value: number | number[]): void {
+  function onUpdateTrim(newTrimValues: number[]): void {
```
beatthat, line 80: ok, let's make this explicit what it is:

```typescript
type StartAndEnd = [number, number]

function onUpdateTrim(startAndEnd: StartAndEnd): void {
```

...and then i think good to go

beatthat (changes requested):

client/src/components/record/video-player.tsx
aaronshiel: Clarified variable and data type in commit: 0e3e873
It would be a scalar if the Slider component was a single value slider with one knob instead of a range slider with 2 knobs.

beatthat (approved):
looks great

PR #67, opened 2021-07-07
page no longer hangs when "record all" pressed on a subject with no category

aaronshiel:


client/src/hooks/graphql/use-with-review-answer-state.tsx
```diff
@@ -109,7 +109,7 @@ export function useWithReviewAnswerState(
           name: subject.name,
           description: subject.description,
           answers: uncategorizedAnswers,
-          recordAll: (status) => recordAnswers(status, subject._id, "none"),
+          recordAll: (status) => recordAnswers(status, subject._id, undefined),
```
beatthat, line 112: whenever possible avoid allowing or using `undefined` explicitly.

In this case, you have a variable that is a string type. You should almost always make the type strictly `string` and not `string | undefined`. Use the empty string to represent `unset`. Javascript makes this easy for you because the empty string is falsy, e.g

```typescript
s = ""
if(!s) { // this is true
}

**Review by beatthat — CHANGES_REQUESTED** [review]

File: client/src/hooks/graphql/use-with-review-answer-state.tsx (hunk 1 above)
Comment (aaronshiel) [reply client/src/hooks/graphql/use-with-review-answer-state.tsx]: 9382971

File: client/src/hooks/graphql/use-with-review-answer-state.tsx [hunk 2]
```diff
@@ -152,15 +152,11 @@ export function useWithReviewAnswerState(
...
-    status: Status,
-    subject: string | undefined,
-    category: string | undefined
-  ) {
+  function recordAnswers(status: Status, subject: string, category: string) {
     navigate(
       `/record?back=${encodeURI(
         `/?subject=${selectedSubject}`
-      )}&status=${status}&subject=${subject || ""}&category=${category || ""}`
+      )}&status=${status}&subject=${subject}&category=${category}`
```
beatthat, line 159: I asked Jaime to add a `urlBuild` helper function in another PR. Once that's in place, we should use it instead of this string-format stuff

beatthat (approved):
looks good

PR #74, opened 2021-07-14
Recommendation button polish

supersweet-dev:
fixed thumbnail check, name and title separated

### DIFF
--- a/client/src/components/my-mentor-card/index.tsx
+++ b/client/src/components/my-mentor-card/index.tsx
@@ -45,7 +45,7 @@ export default function MyMentorCard(props: {
     props.accessToken,
     props.mentor?.thumbnail || ""
   );
-  const thumbnailAvailable = /(ftp|http|https):/.test(thumbnail);
+  const hasThumbnail = Boolean(thumbnail);
   return (
     <div style={{ marginTop: 2, flexGrow: 1, marginLeft: 25, marginRight: 25 }}>
       <Card data-cy="my-mentor-card">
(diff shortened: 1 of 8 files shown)


client/src/components/my-mentor-card/index.tsx
```diff
@@ -45,7 +45,7 @@ export default function MyMentorCard(props: {
     props.accessToken,
     props.mentor?.thumbnail || ""
   );
-  const thumbnailAvailable = thumbnail !== "";
+  const thumbnailAvailable = /(ftp|http|https):/.test(thumbnail);
```
beatthat, line 48: are we sure this shouldn't just be `const thumbnailAvailable = Boolean(thumbnail)`?

How can thumbnail be a non-empty string and be invalid?

client/src/components/my-mentor-card/use-with-recommended-action.tsx
```diff
@@ -145,16 +145,15 @@ export function UseWithRecommendedAction(
   );
   const idleIncomplete = idle?.status === Status.INCOMPLETE;
   const isVideo = mentor?.mentorType === MentorType.VIDEO;
-  const hasThumbnail = mentor?.thumbnail != "";
-
+  const hasThumbnail = /(ftp|http|https):/.test(mentor?.thumbnail);
```
beatthat, line 148: same q: not crazy about this scheme test as opposed to just testing if the string is non-empty...unless we really need it this way?

beatthat (changes requested):
looks good except I don't want to require absolute urls for a valid thumbnail. see inline

beatthat (approved):

PR #75, opened 2021-07-15
center video player, flaky record button fix

aaronshiel:


client/src/hooks/graphql/use-with-data.tsx
```diff
@@ -91,7 +91,7 @@ export function useWithData<T>(fetch: () => Promise<T>): UseData<T> {
       return;
     }
     dispatch({ type: LoadingActionType.SAVING, payload: true });
-    update.action(editedData).catch((err) => {
+    await update.action(editedData).catch((err) => {
```
beatthat, line 94: you can't mix `await` with promise-style `.then().catch()`


Think of `async`/`await` as taking you async code with its callbacks and making it look synchronous and linear.
So if you want to use await, would be like

```typescript
try {
  await update.action(editedData)
}
catch(err) {
  // handle error case
}
```

beatthat (changes requested):
small change

beatthat (changes requested):
let's clean up the shared loading model as discussed in standup

beatthat (changes requested):
just get rid of the error: undefineds

beatthat (approved):

PR #78, opened 2021-07-17
processing text ellipsis animation

aaronshiel:


client/src/hooks/graphql/use-with-upload-status.tsx
```diff
@@ -50,6 +50,8 @@ export function useWithUploadStatus(
   const [uploads, setUploads] = useState<UploadTask[]>([]);
   const [isUploading, setIsUploading] = useState<boolean>(false);
   const [isPolling, setIsPolling] = useState<boolean>(false);
+  const [uploadProcessingText, setUploadProcessingText] =
```
beatthat, line 53: for how we're doing this, maybe cleaner if we just provided a poll count here?

```typescript
const [pollStatusCount, setPollStatusCount]
```

...that way all the stuff about ellipses can be a concert for the presentation component, which seems a cleaner division of responsibility

beatthat (commented):
Looks good. See initial question/comment inline

client/src/hooks/graphql/use-with-upload-status.tsx
aaronshiel: 3632c1d

client/src/hooks/graphql/use-with-upload-status.tsx
```diff
@@ -81,11 +82,11 @@ export function useWithUploadStatus(
     setIsPolling(uploads.some((u) => isTaskPolling(u)));
   }, [uploads]);
 
-  //we start polling because we added an upload to our state when pressed
   useInterval(
     (isCancelled) => {
       fetchUploadTasks(accessToken)
         .then((data) => {
+          setPollStatusCount(pollStatusCount + 1);
```
beatthat, line 89: is `isCancelled` I think no reason to make any additional state updates?

beatthat (changes requested):
looks great. Very small q and possible change request about whether to update anything in the case of `isCancelled` after that, ready to merge

beatthat (approved):

PR #79, opened 2021-07-19
Recommendation button redesign

supersweet-dev:
* Adds isDirty to query to track if mentor has been built with newest answers.
* Recommendation text is above button now, button always says Go.
* Recommendation button now has an icon for each recommendation. 
* Skip loops back to initial suggestion after final recommendation.


client/src/components/home/index.tsx
```diff
@@ -84,6 +85,13 @@ function HomePage(props: {
     startTraining,
   } = useWithReviewAnswerState(props.accessToken, props.search);
 
+  const buildAction = mentor?.isDirty
+    ? startTraining
+    : (mentorId: string) => {
+        const path = `${location.origin}${CLIENT_ENDPOINT}?mentor=${mentorId}`;
```
beatthat, line 91: If `startTraining` is an action provided by `useWithReviewAnswerState`, shouldn't `launchMentor` be an action as well?

client/src/components/my-mentor-card/index.tsx
```diff
@@ -37,6 +37,7 @@ const useStyles = makeStyles(() => ({
 export default function MyMentorCard(props: {
   accessToken: string;
   mentor: Mentor;
+  buildAction: (mentorId: string) => void;
```
beatthat, line 40: seems like mentor is already known and fixed by the caller (since they're passing a mentor).

So seems like `buildAction` should not require (or even accept) a `mentorId` param?

client/src/components/my-mentor-card/recommended-action-button.tsx
```diff
@@ -5,20 +5,29 @@ Permission to use, copy, modify, and distribute this software and its documentat
...
 */
 import { Button, Typography } from "@material-ui/core";
-import { DoubleArrow, PlayCircleFilled } from "@material-ui/icons";
+import { DoubleArrow } from "@material-ui/icons";
 import React from "react";
 import { Mentor } from "types";
 import { UseWithRecommendedAction } from "./use-with-recommended-action";
 export default function RecommendedActionButton(props: {
   setThumbnail: (file: File) => void;
+  buildAction: (mentorId: string) => void;
```
beatthat, line 14: same comment...I suspect `buildAction` should not require or accept a mentor param (because mentor is known and fixed)

beatthat (changes requested):
Looks good. A few comment/questions inline.

The main question is whether we need to parameterize action functions like `build` with a mentor. At the point where that function is being passed into a component, it looks like mentor is already determined, so it would be better to make it captured and locked into the function, no?

client/src/components/home/index.tsx
```diff
@@ -82,8 +82,14 @@ function HomePage(props: {
     selectSubject,
     saveChanges,
     startTraining,
+    launchMentor,
   } = useWithReviewAnswerState(props.accessToken, props.search);
 
+  const buildAction = () =>
```
beatthat, line 88: is there a case for performing any build/launch action if the mentor is undefined?

Maybe this should be

```typescript
const continueAction = () => {
  if(!mentor) {
    console.error("attempt to build or preview a mentor but no mentor set")
    return
  }
  if (mentor.isDirty) {
    startTraining(mentor._id)
  }
  else {
    launchMentor(mentor._id);
  }
}

File: client/src/components/home/index.tsx [hunk 5]
```diff
@@ -166,10 +176,10 @@ function HomePage(props: {
             variant="extended"
             color="primary"
             disabled={!mentor || isTraining || isLoading || isSaving}
-            onClick={() => startTraining(mentor._id)}
+            onClick={buildAction}
             className={classes.fab}
           >
-            Build Mentor
+            {mentor?.isDirty ? "Build Mentor" : "Preview Mentor"}
```
beatthat, line 182: `mentor` has been narrowed above to type `Mentor` (cannot be null or undefined), so everywhere after that narrowing remove the falsy checks on `mentor`

client/src/components/my-mentor-card/use-with-recommended-action.tsx
```diff
@@ -189,22 +215,32 @@ export function UseWithRecommendedAction(
     mentor?.answers.filter((a) => a.status === Status.COMPLETE).length || 0;
   const totalAnswers = mentor?.answers.length || 0;
 
+  return {
+    mentorId: mentor?._id,
```
beatthat, line 219: same issue about narrowing mentor. It would be better to have a check at the top of this function

```typescript
if(!mentor) { // compile-tie types say this is impossible, but if you want to handle it for runtime
  return { ... } // whatever makes sense for the (unexpected) no-mentor case
}

File: client/src/hooks/graphql/use-with-review-answer-state.tsx [hunk 7]
```diff
@@ -313,6 +313,10 @@ export function useWithReviewAnswerState(
         setIsSaving(false);
       });
   }
+  function launchMentor(params: string) {
+    const path = `${location.origin}${CLIENT_ENDPOINT}?mentor=${params}`;
```
beatthat, line 317: use a global url-builder helper function

beatthat (changes requested):
looks good and is close. Biggest issue I see is imprecise narrowing. In general for handling case of something like an undefined mentor object, just check for that condition at the top of a component/function and return accordingly. Everywhere below that check, mentor is defined

beatthat (approved):
looks good

PR #82, opened 2021-07-21
Fix setup bug:

kycarr:
- changing slides too quickly causes slide to get stuck


client/src/hooks/graphql/use-with-setup.tsx
```diff
@@ -189,23 +189,31 @@ export function useWithSetup(
...
   function nextStep() {
-    if (!status || idx === steps.length - 1) {
+    if (!status) {
       return;
     }
     if (isMentorEdited) {
       saveMentorDetails();
     }
-    setIdx(idx + 1);
+    if (idx === steps.length - 1) {
```
beatthat, line 198:
```typescript
function addToIdx(delta: int = 1): void {
  // we have to add steps.length below because stupid js
  // returns negative mods, e.g.
  //    (0 - 1) % 10 == -1 // should be 9
  setIdx((!isNaN(Number(idx)? Number(idx) + (delta + steps.length): 0) % steps.length);
}
```

client/src/hooks/graphql/use-with-setup.tsx
```diff
@@ -189,23 +189,31 @@ export function useWithSetup(
...
   function prevStep() {
-    if (!status || idx === 0) {
+    if (!status) {
       return;
     }
     if (isMentorEdited) {
       saveMentorDetails();
     }
-    setIdx(idx - 1);
+    if (idx === 0) {
```
beatthat, line 212:
```typescript
addToIdx(-1);
```

beatthat (changes requested):
small change request inline about managing +/- index.

Otherwise looks good. I don't love that this force injects 100's of millis of wait time for each `cy.getSettled` call, but I do like that the `getSettled` function is well encapsulated, so we can try to optimize it as cypress improves without having to rewrite the test code

client/src/hooks/graphql/use-with-setup.tsx
```diff
@@ -188,24 +188,35 @@ export function useWithSetup(
     }
   }, [trainError]);
 
+  function addToIdx(delta = 1): void {
+    // we have to add steps.length below because stupid js
+    // returns negative mods, e.g.
+    //    (0 - 1) % 10 == -1 // should be 9
+    setIdx(
```
beatthat, line 195: % is in the wrong place, I probably sent to u incorrectly:

```typescript
setIdx(
      !isNaN(Number(idx))
        ? (Number(idx) + (delta + steps.length)) % steps.length
        : 0
    );
```

beatthat (changes requested):
small error in that `addToIdx` function

beatthat (approved):
looks great!

PR #83, opened 2021-07-23
Redux login config

supersweet-dev:
must fix cypress tests


beatthat (changes requested):
This looks pretty good to me looking at the code.

Trying to test it locally (which you should always do to), and I get hanging loading page pretty much every time i launch /admin

I think the problem is the 5sec timeout on graphql reuqests here:

https://github.com/mentorpal/mentor-admin/blob/29c429fd6fe87f250a5461daf57008bd8b5016f9/client/src/api.ts#L62

...can we temp change this to 30secs just to be able to test locally and confirm there aren't other issues? And also, the current model of having the app hang indefinitely on long requests has no advantages over just letting the requests run longer

PR #84, opened 2021-07-24
Mentor is prompted to fill in transcript if empty

aaronshiel:


beatthat (changes requested):
let's narrow upload-list-item.tsx to use a custom hook that gives its own data.

client/src/types.ts
```diff
@@ -120,6 +120,11 @@ export interface Answer {
   media?: Media[];
 }
 
+export enum AnswerAttentionNeeded {
+  NONE = "NONE",
```
beatthat, line 124: let's never never ever use a string literal value like `NONE` if it's an option to use the empty string

client/src/types.ts
```diff
@@ -167,6 +172,7 @@ export enum QuestionType {
 }
 
 export enum UtteranceName {
+  NONE = "NONE",
```
beatthat, line 175: is this `NONE` coming from the server? I actively hate the idea of a NONE string instead of the empty string. 

If that's how it already works in the server/data, then let's at least limit the scope of handling the NONE string to a single helper function, e.g.

```typescript
export function isNoneUtteranceName(n: UtteranceName): boolean {
  return n && n !== UtteranceName.NONE
}
```

beatthat (changes requested):
Few change requests in line. Biggest thing is I really really dislike the idea of using a magic string like `NONE` when it seems like the empty string would work.

For cases where this is already established and coming from the server, let's make sure to encapsulate that implementation detail into a single helper function like `isNoneUtterance(n: UtteranceName)`

...separate question is about passing the UseRecordState hook as a param to other hooks...wondering if we really need to do this? I can imagine why we might (because if we let each item call `useRecordState` internally they end up w their multiple/distinct copies of some piece of state that should be global/shared)?

Anyway wanted to check on that and make sure...hit me up to discuss

client/src/hooks/graphql/use-with-upload-list-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  function isJobFailed(): boolean {
+    return (
+      jobStatus === UploadStatus.UPLOAD_FAILED ||
+      jobStatus === UploadStatus.TRANSCRIBE_FAILED
+    );
+  }
+
+  const needsAttention = answer?.attentionNeeded ? true : false;
```
beatthat, line 31: this can just be `const needsAttention = Boolean(answer?.attentionNeeded)`

client/src/hooks/graphql/use-with-upload-list-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export interface UseWithUploadListItem {
+  upload: UploadTask;
+  jobStatus: UploadStatus;
+  isJobDone: () => boolean;
+  isJobFailed: () => boolean;
+  cancelling: boolean;
+  needsAttention: boolean;
+  jobTitle: string;
+  gqlPollCount: number;
```
beatthat, line 61: `pollStatusCount` is a better name, because the `gql` is an implementation detail and should not matter

beatthat (changes requested):
looks great. Two tiny change requests and then let's deploy it

beatthat (approved):

PR #85, opened 2021-07-24
Upload system respects that idle videos have no text

aaronshiel:


beatthat (changes requested):
the upload system should determine whether a video is idle on it's own (by checking graphql)

PR #86, opened 2021-07-26
Mentor to redux

supersweet-dev:
Migrates UseWithMentor to Redux, some pages still rely on previous UseWithMentor


client/src/pages/profile.tsx
```diff
@@ -49,31 +52,23 @@ const useStyles = makeStyles(() => ({
...
-  const {
-    editedData: editedMentor,
-    error: mentorError,
-    isLoading: isMentorLoading,
-    isSaving: isMentorSaving,
-    isEdited: isMentorEdited,
-    editData: editMentor,
-    saveMentorDetails,
-  } = useWithMentor(props.accessToken);
+  const { state, saveMentor, editMentor } = useWithMentor(props.accessToken);
```
beatthat, line 55: There really shouldn't be an accessToken available to javascript at all at this point...but if the app is broken in the sense that we still use it even having cookies, the let's isolate it as much as possible.

It's 100% wrong to have components need to know there's such a thing as an accessToken and then need to pass it back to hooks.

Make sure the accessToken is available in redux for the time being and then let `useWithMentor` just get it from `useSelector` until we fully stamp it out

client/src/pages/profile.tsx
```diff
@@ -49,31 +52,23 @@ const useStyles = makeStyles(() => ({
...
+      <NavBar title="My Profile" mentorId={state.editedData?._id} />
       <Paper className={classes.paper}>
         <div className={classes.inputField}>
           <FormControl>
             <InputLabel>Mentor Type</InputLabel>
             <Select
               data-cy="select-chat-type"
               label="Mentor Type"
-              value={editedMentor?.mentorType}
+              value={state.editedData?.mentorType}
```
beatthat, line 71: narrowing above

```typescript
if (!state.editedData) {
  return <div />;
}
```

...means that `state.editedData` CANNOT be undefined here. This is not a compile-time check; this is true at runtime. Make sure to always take proper advantage of narrowing. If you DON'T do the narrowing, you're not actually making you code more safe. Instead, you're making it more confusing/harder to read

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,160 @@
...
+  editedData: undefined,
+  isEdited: false,
+  mentorStatus: MentorStatus.NONE,
+  error: undefined,
+  accessToken: "",
+};
+
+/** Actions */
+
+export const getMentor = createAsyncThunk(
```
beatthat, line 47: redux-toolkit is new to me: do we not need an explicit type on consts created by `createAsyncThunk` because typrescript infers the type? Does vscode show the variable as fully typed on mouseover?

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,160 @@
...
+  "mentor/getMentor",
+  async (headers: Headers) => {
+    return await api.fetchMentor(headers.accessToken);
```
beatthat, line 50: when you return a call to an `async` function, you never need to await, because the return type is `Promise<T>` either way. Usually the linter considers this an error?

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+);
+
+export const saveMentor = createAsyncThunk(
+  "mentor/saveMentor",
+  async (headers: Headers, { rejectWithValue }) => {
```
beatthat, line 56: just as a reader of the code, I would prefer to see the return types at a minimum. Are they difficult to figure out?

client/src/store/slices/mentor/useWithMentor.ts
```diff
@@ -0,0 +1,82 @@
...
+
+interface UseWithMentor {
+  state: mentorActions.MentorState;
+  getMentor: () => void;
+  saveMentor: () => void;
+  saveMentorSubjects: (editedData: Mentor) => void;
+  editMentor: (edits: Partial<Mentor>) => void;
+}
+
+export const useWithMentor = (accessToken: string): UseWithMentor => {
```
beatthat, line 22: get accessToken from the redux state until we don't need it any more (it should be in the cookies?)

beatthat (changes requested):
looks like a good start. See inline comments

client/src/store/slices/mentor/mentorSlice.ts
supersweet-dev: yeah, it seems like it

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export const saveMentor = createAsyncThunk(
+  "mentor/saveMentor",
+  async (
+    headers: { accessToken: string; editedData: Mentor },
+    { rejectWithValue }
+  ) => {
+    try {
+      return api.updateMentorDetails(headers.editedData, headers.accessToken);
+    } catch (err) {
```
beatthat, line 55:
```typescript
// needs await because you're looking for the response in the catch
return await api.updateMentorDetails(headers.editedData, headers.accessToken);
```

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,151 @@
...
+    return api.fetchMentor(accessToken);
+  }
+);
+
+export const saveMentor = createAsyncThunk(
+  "mentor/saveMentor",
+  async (
+    headers: { accessToken: string; editedData: Mentor },
+    { rejectWithValue }
+  ) => {
```
beatthat, line 52: specify return type `Promise<bool>`

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,151 @@
...
+  isEdited: false,
+  mentorStatus: MentorStatus.NONE,
+  error: undefined,
+};
+
+/** Actions */
+
+export const loadMentor = createAsyncThunk(
+  "mentor/loadMentor",
+  async (accessToken: string) => {
```
beatthat, line 42: specify return type

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export const saveMentorSubjects = createAsyncThunk(
+  "mentor/saveMentorSubjects",
+  async (
+    headers: { accessToken: string; editedData: Mentor },
+    { rejectWithValue }
+  ) => {
+    try {
+      return api.updateMentorSubjects(headers.editedData, headers.accessToken);
+    } catch (err) {
```
beatthat, line 70: not sure about this try/catch...I might have led you astray about returning promises?

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+);
+
+export const saveMentorSubjects = createAsyncThunk(
+  "mentor/saveMentorSubjects",
+  async (
+    headers: { accessToken: string; editedData: Mentor },
+    { rejectWithValue }
+  ) => {
```
beatthat, line 67: specify return type

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,151 @@
...
+  LOADING = 1,
+  SAVING = 2,
+  SUCCEEDED = 3,
+  FAILED = 4,
+}
+
+export interface MentorState {
+  data: Mentor | undefined;
+  editedData: Mentor | undefined;
+  isEdited: boolean;
```
beatthat, line 25: `isEditted?: boolean` because all checks of `isEditted` are just truthy

client/src/store/slices/mentor/mentorSlice.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        state.mentorStatus = MentorStatus.LOADING;
+      })
+      .addCase(loadMentor.fulfilled, (state, action) => {
+        state.data = action.payload;
+        state.editedData = action.payload;
+        state.mentorStatus = MentorStatus.SUCCEEDED;
+      })
+      .addCase(loadMentor.rejected, (state) => {
+        state.data = undefined;
```
beatthat, line 107:
```typescript
del state.data  // or is it del state['data']
```

client/src/store/slices/mentor/useActiveMentor.ts
```diff
@@ -0,0 +1,86 @@
...
+  const loginState = useAppSelector((state) => state.login);
+
+  useEffect(() => {
+    if (state.data) {
+      return;
+    }
+    if (loginState.accessToken) {
+      loadMentor();
+    }
+  }, []);
```
beatthat, line 34:
```typescript
// what if we got here but the user wasn't logged in yet, but then they subsequently did login? so...
}, [loginState]);
```

client/src/store/slices/mentor/useActiveMentor.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const loadMentor = () => {
+    if (
+      state.mentorStatus === mentorActions.MentorStatus.NONE ||
+      state.mentorStatus === mentorActions.MentorStatus.FAILED
+    ) {
+      if (loginState.accessToken) {
```
beatthat, line 41: this makes me nervous. Can we dispatch an error? e.g.

```typescript
if (loginState.accessToken) {
  dispatch({ 
    type: MENTOR_LOAD_FAILED, 
    payload: "cannot load mentor while unauthenticated"
  })
  return
}
```

beatthat (changes requested):
looks good changes inline

beatthat (approved):

PR #87, opened 2021-07-27
Import mentor

kycarr:


client/src/components/import-export/import-subject.tsx
```diff
@@ -0,0 +1,271 @@
...
+  subjects: Subject[];
+  mapSubject: (curSubject: Subject, newSubject: Subject) => void;
+}): JSX.Element {
+  const classes = useStyles();
+  const [isExpanded, setIsExpanded] = useState(false);
+  const [subjectSearch, setSubjectSearch] = useState<Subject>();
+  const { preview, subjects, mapSubject } = props;
+  const { editType, importData: subject, curData: curSubject } = preview;
+
+  if (subject === undefined && curSubject === undefined) {
```
beatthat, line 66: why not `if(!subject && !curSubject)` ?

client/src/components/import-export/import-subject.tsx
```diff
@@ -0,0 +1,271 @@
...
+    return <div />;
+  }
+
+  const categories: ImportPreview<Category>[] = [];
+  subject?.categories?.forEach((c) => {
```
beatthat, line 71: is this not just a `map` (one to one)?

client/src/components/import-export/import-view.tsx
```diff
@@ -0,0 +1,141 @@
...
+    importPreview,
+    cancelImport,
+    confirmImport,
+    mapSubject,
+    mapQuestion,
+  } = props.useImportExport;
+  const { data: subjects } = useWithSubjects();
+  const { data: questions } = useWithQuestions();
+
+  if (importedJson === undefined || importPreview === undefined) {
```
beatthat, line 59: `if(!importedJson && !importPreview)` ?

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const [importedJson, setImportJson] = useState<MentorExportJson>();
+  const [importPreview, setImportPreview] = useState<MentorImportPreview>();
+  const { mentorState } = useActiveMentor();
+  const mentor = mentorState.data;
+
+  function exportMentor() {
+    if (!mentor) {
+      return;
+    }
+    api.exportMentor(mentor._id).then((m) => {
```
beatthat, line 40: any reason this shouldn't just use `async/await`?

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -0,0 +1,157 @@
...
+      const element = document.createElement("a");
+      element.setAttribute(
```
beatthat, line 42: can we avoid the dom manipulation here and do something more `React`-ish? e.g. can this be handled in a use effect? Beyond avoid jquery, I think we should strenuously avoid ever having a hook manipulate the dom as a violation of separation of concerns (hooks manage data, components manage presenation/dom)

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  function mapQuestion(question: Question, replacement: Question) {
+    if (!importedJson || !importPreview || !mentor) {
+      return;
+    }
+    const json = { ...importedJson };
+    // if the replacement question is already referenced elsewhere in the import, remove it so we don't include it twice
+    let rIdx = json.questions.findIndex((q) => q._id === replacement._id);
+    ~rIdx && json.questions.splice(rIdx, 1);
```
beatthat, line 118: all these splices of `json.question` are mutations of state, no?

We need to never mutate the existing state. If you want to simplify and void wasteful/repetitive copying, then just copy one time above, e.g.

```typescript
const json = { 
  ...importedJson,
  questions: ...json?.questions || []
};
```

cypress/cypress/downloads/mentor.json
```diff
@@ -0,0 +1 @@
+undefined
```
beatthat, line 1: we need this as a file? maybe call it `mentor-undefined.json` if we really do this

beatthat (changes requested):
This looks really good. 

there's a lot to go through here. Biggest initial response is whether we really need `mentorType` as a property of question and can we move `minVideoLength` to a generic list of properties?

Other initial comments inline

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -0,0 +1,157 @@
...
+  function cancelImport() {
+    setImportJson(undefined);
+    setImportPreview(undefined);
+  }
+
+  function mapSubject(subject: Subject, replacement: Subject) {
+    if (!importedJson || !importPreview || !mentor) {
+      return;
+    }
+    const json = { ...importedJson };
```
beatthat, line 98:
```typescript
const json = {
   ...importedJson,
  subjects: ...importedJson.subjects || []
}
```

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -0,0 +1,157 @@
...
+      return;
+    }
+    setImportJson(json);
+    api.importMentorPreview(mentor._id, json).then((p) => {
+      console.log(JSON.stringify(p, null, " "));
+      setImportPreview(p);
+    });
+  }
+
+  function confirmImport() {
```
beatthat, line 79:
```typescript
function onConfirmImport(): void {
  api.importMentor(mentor._id, importedJson, accessToken)
 .then(() => {})
 .catch(() => {}) // super important! if you call a function from a ui event w promises u need to handle rejections
}
```

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -0,0 +1,157 @@
...
+    reader.onload = function (e) {
+      if (typeof e.target?.result === "string") {
+        const json = JSON.parse(e.target?.result);
+        updateImport(json);
+      }
+    };
+    reader.readAsText(file);
+  }
+
+  function updateImport(json: MentorExportJson) {
```
beatthat, line 68: make `async` and use `await`

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -0,0 +1,157 @@
...
+      );
+      element.setAttribute("download", "mentor.json");
+      element.style.display = "none";
+      document.body.appendChild(element);
+      element.click();
+      document.body.removeChild(element);
+    });
+  }
+
+  function importMentor(file: File) {
```
beatthat, line 54: return type `void`

beatthat (changes requested):
inline requests

PR #92, opened 2021-07-28
All urls use builder

supersweet-dev:


client/src/components/setup/record-idle-slide.tsx
```diff
@@ -20,7 +21,10 @@ export function RecordIdleSlide(props: {
 
   function onRecord() {
     navigate(
-      `/record?videoId=${idle.question._id}&back=${encodeURI(`/setup?i=${i}`)}`
+      urlBuild("/record", {
+        videoId: idle.question._id,
+        back: encodeURI(`/setup?i=${i}`),
```
beatthat, line 26: trying to think this through: do we need to encode this param here (and it gets encoded another time by `urlBuild` right?)?

client/src/hooks/graphql/use-with-review-answer-state.tsx
```diff
@@ -187,17 +187,21 @@ export function useWithReviewAnswerState(
...
   function recordAnswers(status: Status, subject: string, category: string) {
     navigate(
-      `/record?back=${encodeURI(
-        `/?subject=${selectedSubject}`
-      )}&status=${status}&subject=${subject}&category=${category}`
+      urlBuild("/record", {
+        status: status,
+        subject: subject,
+        category: category,
+        back: encodeURI(`/?subject=${selectedSubject || ""}`),
```
beatthat, line 194: if there's no `selectedSubject`, would it not be better to exclude the param altogether?

```typescript
back: encodeURI(urlBuild('/', selectedSubject? { subject: selectedSubject } : {}),
```

beatthat (changes requested):
...all the places where we're passing params to `urlBuild` where the param is itself a url, should we not just use `urlBuild` to generate those urls and NOT do the explicit `encodeURI`?

beatthat:
Let's fix `urlBuilder` itself so that

 - [ ] urlBuilder handles uriEncoding of query params (I suspect it already does, but let's verify)
 - [ ] urlBuild does NOT include empty param by default

...something like

```typescript
interface UrlBuildOpts {
  includeEmptyParams?: boolean
}
export function urlBuild(base: string, params: Record<string, string>, opts: UrlBuildOpts): string {
  const query = new URLSearchParams();
  Object.keys(params).forEach((n) => {
    if ((params == null || typeof params === 'undefined') && !opts?.includeEmptyParams
    /**
      * we have to distinguish between params where the value is `undefined` or `null`
      * (which we generally want to exclude) vs. params where the value is falsy but meaningful (e.g. `0`)
      */
    const pval = (params[n] !== null && typeof params !== 'undefined') 
      ? params[n]: ""
    if(!(pval || opts?.includeEmptyParams) {
      return
    }
    query.append(n, encodeURL(pval));
  }
  const qs = query.toString()
  return (qs)? `${base}?${qs}` : base;  // don't put ? if no query string
  return `${base}?${query.toString()}`;
}
```

cypress/cypress/integration/setup.spec.ts
```diff
@@ -539,7 +542,7 @@ describe("Setup", () => {
     });
   });
 
-  it("shows required subject, repeat after me, questions slide", () => {
+  it.only("shows required subject, repeat after me, questions slide", () => {
```
beatthat, line 545: make sure to delete any `.only`

beatthat (changes requested):
other note: be careful there are no `.only`

beatthat (approved):
looks good, but make sure to test locally in integrated app before merge to main

PR #95, opened 2021-08-03
Gha generates snapshots

aaronshiel:
Let's use some test branches from this branch to confirm that it works:

 - [x] generates-new-snapshots - a branch with new snapshot calls in cypress and confirm that GHA generates and commits the new snaps
 - [x] fails-on-snapshot-mismatch - a branch where maybe two existing snap images have been changed manually to fail and test that...A) the tests fails, and B) the snapshot diffs for just the failed snaps are published to artifacts


.github/workflows/test-and-publish.yml
```diff
@@ -82,4 +96,4 @@ jobs:
         if: github.ref == 'refs/heads/main'
         run: docker tag "uscictdocker/mentor-admin:${{ steps.tag.outputs.image_tag }}" "uscictdocker/mentor-admin:latest"
       - name: Docker push
-        run: docker push --all-tags uscictdocker/mentor-admin
+        run: docker push --all-tags uscictdocker/mentor-admin
```
beatthat, line 99: newline at EOF

beatthat (commented):
looks cool! one small fix for EOF

Have you tested this w some scenarios (unfortunately, can't think how to automate tests)

 - [ ] detects / adds / commits any missing snapshots on push
 - [ ] creates / commits ALL snapshots if snapshots directory completely deleted
 - [ ] commit FAILS (as it used to) if an existing/commited snapshot no longer matches (e.g.just add a red square to a snapshot)
 - [ ] user can review diffs of failed snapshots in artifacts
 - [ ] user can update a single failing snapshot by delete, commit, push (this is a variation of 'detects / adds / commits any missing snapshots on push')

.github/workflows/test-and-publish.yml
```diff
@@ -33,15 +33,29 @@ jobs:
...
+        if: steps.set-has-new-snapshots.outputs.has_new_snapshots == 'true'
+        run: |
+          git config user.name ${{ github.actor }}
+          git config user.email github-actions@github.com
+          git add $(git ls-files --others --exclude-standard -- cypress/cypress/snapshots)
+          git commit -m "gha, adding new snapshots"
+          git push
       - name: If Cypress failed, retrieve image diffs
-        run: make test-e2e-image-snapshots-copy && mkdir -p client/cypress/snapshots/snapshot_failed_diffs
+        run:  make test-e2e-copy-failed_diffs
```
beatthat, line 51: looks like there is an extra whitespace here before `make` which is a tiny thing but delete it if that's the case

beatthat (changes requested):
this looks fantastic! Fix the 2 tiny format issues mentioned inline and then let's merge

beatthat (approved):

PR #96, opened 2021-08-03
If setup is not complete, redirect to setup page from home page:

kycarr:
- if no mentor info, redirect
- otherwise, offer to redirect and go to step that needs completing

### DIFF
--- a/client/src/hooks/graphql/use-with-setup.tsx
+++ b/client/src/hooks/graphql/use-with-setup.tsx
@@ -105,6 +105,7 @@ export function useWithSetup(search?: { i?: string }): UseWithSetup {
     if (!mentor || isMentorSaving || isMentorLoading || !isConfigLoaded()) {
       return;
     }
+    console.log(`n=${mentor.name} fn=${mentor.firstName} t=${mentor.title}`);
     const isMentorInfoDone = Boolean(
       mentor.name && mentor.firstName && mentor.title
     );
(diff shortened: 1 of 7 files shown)


client/src/hooks/graphql/use-with-setup.tsx
```diff
@@ -110,6 +110,7 @@ export function useWithSetup(
     if (!mentor || isMentorSaving || isMentorLoading || !isConfigLoaded()) {
       return;
     }
+    console.log(`n=${mentor.name} fn=${mentor.firstName} t=${mentor.title}`);
```
beatthat, line 113: let's get rid of the `console.log`

beatthat (changes requested):
Get rid of the `console.log`

Then only other question:

In `use-with-setup`...we will be either replacing `use-w-mentor` w the redux version or converting `use-w-mentor` to be a redux adapter in that other PR, right?

beatthat (changes requested):
We should cleanup some things about this the setup dialog

- [x] where in setup the user goes to should be determined by the use-setup hook, and not the component, e.g. with a function like `navigateToMissingSetup`
- [x] anywhere we generate urls with query strings, use the `urlBuild` helper function
- [ ] for generating setup urls in particular, we need to stop using indexes and use something like an enum instead (see example below)
- [ ] (optional) would be better if SetupMissingDialog were an independent component / tsx file, ideally it should live under /setup and not home


```typescript
// in useSetup
export enum SetupScreens {
  MentorInfo = 1,
  ...
}

function navigateToSetup(screen: SetupScreen?): void {
  if(screen) {
    urlBuild("/setup", { i: screen })
  }
  else {
    navigate("/setup");
  }
}

PR #104, opened 2021-08-05
thumbnail managed by mentor slice

supersweet-dev:
removes usewiththumbnail
thumbnail is uploaded and fetched through mentor state


client/src/components/my-mentor-card/index.tsx
```diff
@@ -54,18 +53,13 @@ export default function MyMentorCard(props: {
   accessToken: string;
   continueAction: () => void;
 }): JSX.Element {
-  const { mentorState, editMentor } = useActiveMentor();
+  const { mentorState, editMentor, saveThumbnail } = useActiveMentor();
```
beatthat, line 56: don't do this. Do this:

```typescript
const thumbnail = useSelector<State, string>(s => {
    // selectActiveMentor can be a helper function exported from useActiveMentor or similar
   selectActiveMentor(s)?.thumbnail || " 
})
```

The reason has to do with scope of selectedState and how that impacts the triggering of render calls

client/src/store/slices/mentor/index.ts
```diff
@@ -57,6 +57,25 @@ export const saveMentor = createAsyncThunk(
...
+export const saveThumbnail = createAsyncThunk(
+  "mentor/saveThumbnail",
+  async (headers: {
+    accessToken: string;
+    mentorId: string;
+    file: File;
+  }): Promise<string | unknown> => {
+    try {
+      return await api
+        .uploadThumbnail(headers.mentorId, headers.file)
```
beatthat, line 69: this is just silly to save and then fetch here. Can `uploadThumbnail` return the thumbnail url?

beatthat (changes requested):
getting close

Main issue is `useThumbnail` should not have any mentor state. See comments inline. Also we need to just change GQL so that updateThumbnail can return the thumbnail url

PR #105, opened 2021-08-05
Fix: video looping bug

aaronshiel:


aaronshiel:
@beatthat Should probably wait till next week to deploy this one. I tested it as much as I could and it seems fine, but 99% chance the bug won't even occur and if it does, won't prevent the user from moving on or anything.

PR #109, opened 2021-08-06
Thumbnail in store rebased

beatthat:
These should be the changes we worked on together, rebased onto careerfair-integration branch


client/src/api.ts
```diff
@@ -831,7 +831,7 @@ export async function uploadThumbnail(
       "Content-Type": "multipart/form-data",
     },
   });
-  return getDataFromAxiosResponse(result, []);
+  return getDataFromAxiosResponse(result, ["thumbnail"]);
```
beatthat, line 834: this probably needs to be `['data', 'thumbnail']` (because this request isn't actually graphql)?

beatthat (commented):
looks good...am curious about that dataPath (question inline)

Otherwise if works locally, let's merge and deploy on monday

client/src/api.ts
supersweet-dev: I actually as that and couldn't figure why it wasn't working but changing it to just "thumbnail" fixed it somehow
the actual response has the shape
data.data.thumbnail

beatthat: The json in the response is just 

```json
{ "data": { "thumbnail": "" } }
```

...the first `data` property you're referring to is a `AxiosResponse` object

kycarr (approved):

PR #115, opened 2021-08-11
Dropdown Button component, applied to subjects page

aaronshiel:


beatthat (commented):
ah ok...so this component doesn't exist in MaterialUI and you had to build the whole thing? Let's just look at it together tmrw before merge

benjamid:
Hey guys, I have a slight concern here from monitoring this. I am pretty sure there *is* a material UI component for this button type: https://material-ui.com/components/button-group/ (seems to be going by "Split Button" in that terminology). While it might involve tossing some work, we would save in maintenance if we can use something already there.

aaronshiel:
> Hey guys, I have a slight concern here from monitoring this. I am pretty sure there _is_ a material UI component for this button type: https://material-ui.com/components/button-group/ (seems to be going by "Split Button" in that terminology). While it might involve tossing some work, we would save in maintenance if we can use something already there.

I don't believe SplitButton is an actual material UI component. On that page, it's showing how a ButtonGroup can be _used_ to create a SplitButton. (Which it provides the code for, which I used as my starting point)

benjamid:
Gotcha, well in that case it sounds like we might be doing the best we can then. Surprising that there's not an off-the-shelf component but such is life I guess.

________________________________

PR #120, opened 2021-08-16
separate use[Active]Mentor (redux) from useMentorEdits (useState hook)

kycarr:


client/src/store/slices/mentor/useActiveMentor.ts
```diff
@@ -0,0 +1,86 @@
...
+*/
+import { useEffect } from "react";
+import { useDispatch } from "react-redux";
+import { Mentor } from "types";
+import { LoadingError } from "hooks/graphql/loading-reducer";
+import { useAppSelector } from "store/hooks";
+import * as mentorActions from ".";
+import { RootState } from "store/store";
+
+export interface UseActiveMentor {
```
beatthat, line 15: let's change all the pre-selected data here to functions, e.g.

```typescript
interface SelectFromMentorStateFunc<T> {
  (mentorState: MentorState) => T
}

export interface UseActiveMentor {
  // mentor?: Mentor;
  mentorSelect<T>: (selector: SelectFromMentorStateFunc<T>) => T;
  //mentorError?: LoadingError;
  mentorError: () => LoadingError | null;
  // mentorStatus: mentorActions.MentorStatus;
  mentorStatus: () => mentorActions.MentorStatus;
  // isMentorLoading: boolean;
  isMentorLoading: () => boolean;
  loadMentor: (mentorId?: string) => void;
  // onMentorUpdated should NOT exist,
  // it is inherently provided by any user who selects mentor data
  // onMentorUpdated: (mentor: Mentor) => void;
  clearMentorError: () => void;
}
````

**Review by beatthat — CHANGES_REQUESTED** [review]
looks good! Main change request is about how redux mentor data is provided. Should be all via lazy-load functions (to allow narrowing of dependencies and minimizing of render callbacks)

File: client/src/components/home/index.tsx [hunk 2]
```diff
@@ -93,7 +96,10 @@ function HomePage(props: {
     startTraining,
   } = useWithReviewAnswerState(props.accessToken, props.search);
   const defaultMentor = props.user.defaultMentor._id;
-  const { mentor, isMentorEdited, loadMentor } = useMentor;
+
```
beatthat, line 99: We don't need to select the whole active mentor here, I think we just need something like this:

```typescript
  React.useEffect(() => {
    if (!setupStatus || !showSetupAlert) {
      return;
    }
    setShowSetupAlert(!setupStatus.isSetupComplete);
  }, [setupStatus]);
  const mentorId = useActiveMentor(m => m.data?._id || "");
  if (!(mentorId && setupStatus)) {
    return (
      <div>
        <NavBar title="Mentor Studio" mentorId={""} />
        <CircularProgress />
      </div>
    );
  }
  const mentorIsDirty = useActiveMentor(m => m.data.isDirty);
  const mentorSubjectNamesById: Record<string, string> = useActiveMentor(
    m => m.data.subjects.reduce((acc: Record<string, string>, cur: Subject) => {
      acc[cur._id] = cur.name;
      return acc
    }, {})
  );
```

client/src/components/my-mentor-card/index.tsx
```diff
@@ -52,16 +56,13 @@ const useStyles = makeStyles(() => ({
...
 }): JSX.Element {
-  const {
-    mentor,
-    isMentorLoading,
-    isMentorSaving,
-    mentorError,
-    editedMentor,
-    editMentor,
-  } = props.useMentor;
+  const mentor = useActiveMentor((state) => state.data);
```
beatthat, line 61: you don't need the whole mentor here. In fact, you already have a function that you should use as a selector...


```typescript
  /**
  const mentor = useActiveMentor((state) => state.data);
  const mentorError = useActiveMentor((state) => state.error);
  // generally defer calling any of these isActiveMentorXXX
  // functions until the place you actually need the result
  const isMentorLoading = isActiveMentorLoading();
  const isMentorSaving = isActiveMentorSaving();
  const { editedMentor, editMentor } = props.useMentor;

  if (!mentor || !editedMentor) {
    return <div />;
  }
  const mentorInfo = parseMentor(mentor);
  */
  const mentorId = useActiveMentor(ms => ms.data?._id || "");
  if (!(mentorId && editedMentor)) {
    return <div />;
  }
  const mentorInfo = useActiveMentor(ms => parseMentor(ms.data));
```

client/src/hooks/graphql/use-with-import-export.tsx
```diff
@@ -32,7 +35,8 @@ export function useWithImportExport(accessToken: string): UseWithImportExport {
   const [importedJson, setImportJson] = useState<MentorExportJson>();
   const [importPreview, setImportPreview] = useState<MentorImportPreview>();
   const [isUpdating, setIsUpdating] = useState(false);
-  const { mentor, loadMentor } = useWithMentor();
+  const mentor = useActiveMentor((state) => state.data);
```
beatthat, line 38: make sure to only select what you're using as in other examples (didn't check, maybe in this case you really use the whole mentor?)

client/src/hooks/graphql/use-with-record-state.tsx
```diff
@@ -58,8 +62,12 @@ export function useWithRecordState(
   const [followUpQuestions, setFollowUpQuestions] = useState<string[]>([]);
   const [curAnswer, setCurAnswer] = useState<CurAnswerState>();
   const pollingInterval = parseInt(filter.poll || "");
-  const { mentor, mentorError, isMentorLoading, loadMentor, clearMentorError } =
-    useWithMentor();
+
+  const mentor = useActiveMentor((state) => state.data);
```
beatthat, line 66: only select what you will use

client/src/hooks/graphql/use-with-record-state.tsx
```diff
@@ -58,8 +62,12 @@ export function useWithRecordState(
...
+  const mentorError = useActiveMentor((state) => state.error);
+  const isMentorLoading = isActiveMentorLoading();
```
beatthat, line 68: defer this function call until result is needed

beatthat (changes requested):
This is good. It needs at least one more pass to clean up with respect the the following super import react/redux concept: 

Select only what you're actually going to use. DO NOT, select something larger than what you need and then pick out the pieces you care about later in code.

There are multiple reasons for this but the most important is that we want to minimize re-renders, and React/Redux will rerender a component any time the *OUTPUT* of a `useSelect` function changes (subsequent to a dispatch that was processed through the redux store)

So I put some examples inline in the changes. We should apply the same approach in all files that `useActiveMentor`

beatthat (commented):
This looks good to me. @kycarr can you take a look before we merge and just make sure we're not creating an unreasonable amount of conflict with the changes you're making for `useQuestions`?

beatthat (approved):

PR #122, opened 2021-08-17
Login owns mentor data

supersweet-dev:
Login State holds active mentor


beatthat (commented):
Let's complete and merge the `useActiveMentor` PR first and then review this one. It looks to me like this is including a lot of the same changes from the other PR?

beatthat (changes requested):
@supersweet-dev can you either merge the target branch back here or rebase and then we'll review?

beatthat (commented):
after we get main merged back to the integration branch, hit me up and let's walk through this. What I had been hoping for was that we could essentially store all data related to a login's mentors nested under `login` but still treat it as mentor data. This would be straightfoward to do (I think?) in the non reduxtoolkit world, but maybe slices make it tricky? I just want to understand the new setup

PR #123, opened 2021-08-18
Merged with main

supersweet-dev:


beatthat (changes requested):
still not passing all tests...need other eyes to fix?

supersweet-dev:
> still not passing all tests...need other eyes to fix?

yeah, cypress tests are flaky all over and I'm getting query syntax errors when testing locally

beatthat (changes requested):
looks good...maybe a few tiny changes to be cleaner in cypress expectations (see inline). After that I vote merge

beatthat (approved):

PR #152, opened 2021-11-05
Final page in setup does not ask to build

aaronshiel:
It still prompts you to complete your setup if it's incomplete, but if your setup is complete, guides you to the My Mentor page


client/src/components/setup/welcome-slide.tsx
```diff
@@ -16,7 +16,7 @@ export function WelcomeSlide(props: {
   return (
     <Slide
       classes={classes}
-      title="Welcome to MentorPal!"
+      title="Welcome to MentorStudio!"
```
kycarr, line 19: Are we changing the name of the admin portal to Mentor Studio? Are there other places that also need changing in other projects?

aaronshiel:
Yeah, this was per Ben’s request

client/src/components/setup/welcome-slide.tsx
aaronshiel: eah, this was per Ben’s request

PR #167, opened 2022-02-23
Transcript fix

aaronshiel:
After an upload completes, the local transcript only gets updated if the upload task resulted in a new transcript (i.e. not null)


client/src/hooks/graphql/use-with-record-state.tsx
```diff
@@ -244,17 +244,21 @@ export function useWithRecordState(
     const idx = answers.findIndex((a) => a.answer.question === upload.question);
     if (idx !== -1) {
       const answer = answers[idx];
+      const newTranscript =
+        upload.transcript || upload.transcript === ""
```
kycarr, line 248: I think can do upload.transcript !== undefined for the same effect (empty string or string).

aaronshiel: [af4e67d](https://github.com/mentorpal/mentor-admin/pull/167/commits/af4e67de2ce46a8835a72f9f6d95d0df42c6855b)

kycarr (approved):

PR #190, opened 2022-05-18
@sentry/react

aaronshiel:
Reports all uncaught exceptions to sentry, and samples 10% of transactions.


client/src/pages/index.tsx
```diff
@@ -10,9 +10,11 @@ import LoginPage from "components/login";
 import HomePage from "components/home";
 import { useWithLogin } from "store/slices/login/useWithLogin";
 import { LoginStatus } from "store/slices/login";
+import { loadSentry } from "helpers";
 
 function IndexPage(): JSX.Element {
   const { state: loginState } = useWithLogin();
+  loadSentry();
```
mbsimonovic, line 17: i'm not sure this is the place to load sentry, perhaps Kayla or Dilan can provide insight.

Ideally i'd like to have sentry as early as possible, ideally as soon as the html is parsed and first script tag evaluated. But that's not as convenient to configure as this react integration. So can you make sure that sentry init is called before any other react logic kicks in.

aaronshiel: I wonder if it's just better to go with the gatsby sentry config. With that, sentry is added as a plugin and is loaded very early in the [bootstrapping phase](https://www.gatsbyjs.com/docs/conceptual/overview-of-the-gatsby-build-process/#:~:text=gatsby%20develop%20will%20run%20at,up%20to%20date%20results%20already.).

mbsimonovic: could be, if you have time you can try both. does gatsby use react under the hood?

aaronshiel: yes gatsby does

aaronshiel:
According to [the gatsby lifecycle docs](https://www.gatsbyjs.com/docs/conceptual/gatsby-lifecycle-apis/), gatsby-browser.js is setup early in the bootstrapping process. Initializing sentry there works too

client/gatsby-browser.js
```diff
@@ -13,6 +13,9 @@ import "videojs-record/dist/css/videojs.record.css";
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import Record from "videojs-record/dist/videojs.record.js";
 import "styles/layout.css";
+import { loadSentry } from "./src/helpers";
+
+loadSentry();
```
mbsimonovic, line 18: is it disabled when developing locally?

client/src/helpers.ts
```diff
@@ -95,3 +97,14 @@ export function onTextInputChanged(
...
 }
+
+export function loadSentry(): void {
+  Sentry.init({
+    dsn: "https://d137124c5ac546639e2536f860a92798@o1081855.ingest.sentry.io/6419221",
+    integrations: [new BrowserTracing()],
+
+    // We recommend adjusting this value in production, or using tracesSampler
+    // for finer control
+    tracesSampleRate: 0.1,
```
mbsimonovic, line 108: how about 0 for v2 and say 0.2 for careerfair?

aaronshiel: Well that would incur the previous issue with gatsby not knowing which stage it's on, the same reason we put the lambda endpoints in the config. So I guess it would be helpful to also put a generic "server" field in the DB config that could be "v2" or "cf" for use cases such as this one?

client/gatsby-browser.js
aaronshiel: [8ab6a80](https://github.com/mentorpal/mentor-admin/pull/190/commits/8ab6a80d8031b2baf19c39fe1ed81595cf5beaa1)

client/src/helpers.ts
mbsimonovic: not anymore, the new cicd pipeline builds separately for each stage and can pass the stage parameter

aaronshiel: how does this look: [e1bc691](https://github.com/mentorpal/mentor-admin/pull/190/commits/e1bc691abd5543ad375a3ed414f14b36df13dc91)

mbsimonovic (approved):

PR #201, opened 2022-06-03
User Feedback panel should not recommend unanswered questions

cassgtz:


client/src/pages/feedback.tsx
```diff
@@ -129,7 +140,7 @@ function FeedbackItem(props: {
     await updateUserQuestion(feedback._id, answerId || "");
     onUpdated();
   }
-
+  console.log(GetCompleteAnswers(mentorAnswers || []));
```
aaronshiel, line 143: remove log

client/src/pages/feedback.tsx
```diff
@@ -116,6 +116,17 @@ const columnHeaders: ColumnDef[] = [
   },
 ];
 
+function GetCompleteAnswers(mentorAnswers: Answer[]) {
```
aaronshiel, line 119: This works totally fine, but this whole function can be achieved in a single line using the filter function. If you have the time, give it a shot. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

aaronshiel (changes requested):

client/src/pages/feedback.tsx
```diff
@@ -176,7 +176,11 @@ function FeedbackItem(props: {
             <Autocomplete
               key={`${feedback._id}-${feedback.updatedAt}`}
               data-cy="select-answer"
-              options={mentorAnswers || []}
+              options={
+                mentorAnswers?.filter(
+                  (mentorAnswer) => mentorAnswer.status == "COMPLETE"
```
aaronshiel, line 181: Instead of comparing to the literal string "COMPLETE", you should compare it to the enumerated value type for status, something like Status.COMPLETE

aaronshiel (changes requested):

client/src/pages/feedback.tsx
aaronshiel: Enumerators are data structures of constant length that hold a set of constant values

aaronshiel (approved):
Good Job! 🎉  🎉

PR #203, opened 2022-06-07
Skip button functionality improvements

lwoluke:
In order to finish the development for the subtasks, I needed to revamp the skip functionality so that the recommendations are in an array. This allowed for the skip button to loop back to the start. Originally, the skip button would only store the information after it to skip to, so once it gets to the end then there isn’t anything that would happen. Also, the "skip button" was actually a link object that would update the recommendation (as a button should be doing), making this also a poor design choice. I fixed this by making this an actual button. I also added functionality so that when there was only 1 element in the recommendations list, the skip button was deactivated meaning that nothing would happen when the user would click on this and it would be grayed out. After this, I changed the styles of the button for better UX. I also fixed duplicate code for the recommended action button.

If you want to learn more, click [this link](https://docs.google.com/document/d/1I5ktKDujmBZA4xDcDCWmdAvKn5eoOZjoxGRATk996zI/edit)


aaronshiel (approved):
Good job! 🎉

PR #206, opened 2022-06-14
User feedback

aaronshiel:


cassgtz:
not sure art my naming conventions!

cassgtz:
not sure abt my naming conventions!

client/src/pages/feedback.tsx
```diff
@@ -116,16 +128,70 @@ const columnHeaders: ColumnDef[] = [
...
+    accessToken,
+    feedback,
+    mentorAnswers,
+    mentorQuestions,
+    onUpdated,
+    queueList,
+    setQueueList,
+  } = props;
+  const [currentStatus, setcurrentStatus] = React.useState<Status>(); // for disabling/enabling queue button
+  const [currentID, setCurrentID] = React.useState<string>(); // grab ID of the selected option
```
aaronshiel, line 181: maybe "selectedAnswerId" would be a better name?

client/src/pages/feedback.tsx
```diff
@@ -116,16 +128,70 @@ const columnHeaders: ColumnDef[] = [
...
+
+  // function to add/remove from queue
+  async function handleClick(currentID: string, accessToken: string) {
```
aaronshiel, line 184: queueButtonClicked

client/src/pages/feedback.tsx
```diff
@@ -116,16 +128,70 @@ const columnHeaders: ColumnDef[] = [
   },
 ];
 
+function FormatMentorQu(
```
aaronshiel, line 131: It is convention that functions are lowerCamelCase, so this should be formatMentorQu (why not formatMentorQuestions?)

client/src/pages/feedback.tsx
```diff
@@ -190,7 +257,14 @@ function FeedbackItem(props: {
                 flexGrow: 1,
               }}
               renderOption={(option) => (
-                <Typography align="left">
+                <Typography
+                  style={{
+                    color:
+                      option.status == Status.INCOMPLETE ? "grey" : "black",
```
aaronshiel, line 263: use === instead of ==
This explains the difference: https://howtodoinjava.com/typescript/equals-vs-strict-equals/

cypress/cypress/integration/feedback.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+Permission to use, copy, modify, and distribute this software and its documentation for educational, research and non-profit purposes, without fee, and without a written agreement is hereby granted, provided that the above copyright notice and subject to the full license file found in the root of this software deliverable. Permission to make commercial use of this software may be obtained by contacting:  USC Stevens Center for Innovation University of Southern California 1150 S. Olive Street, Suite 2300, Los Angeles, CA 90115, USA Email: accounting@stevens.usc.edu
+
+The full terms of this copyright and license should always be found in the root directory of this software deliverable as "license.txt" and if these terms are not found with this software, please contact the USC Stevens Center for the full license.
+*/
+import { cySetup, cyMockDefault, mockGQL } from "../support/functions";
+import mentor from "../fixtures/mentor/clint_new";
+import { feedback as userQuestions } from "../fixtures/feedback/feedback";
+
+describe("Feedback", () => {
+  it("Format dropdown options", () => {
```
aaronshiel, line 12: description should be something like "dropdown un-recorded questions are greyed out", this test should also be moved to record-queue.spec.ts

cypress/cypress/integration/record-queue.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  cyMockDefault,
+  mockGQL,
+  cyMockFollowUpQuestions,
+} from "../support/functions";
+import { feedback as userQuestions } from "../fixtures/feedback/feedback";
+import mentor from "../fixtures/mentor/clint_new";
+
+describe("Mentor Record Queue", () => {
+  describe("Feedback Page", () => {
+    it.only("Testing record queue on feedback page", () => {
```
aaronshiel, line 17: remove this test since it doesn't test anything

client/src/pages/feedback.tsx
cassgtz: should currentStatus also be changed to selectedAnswerStatus then?

cypress/cypress/integration/record-queue.spec.ts
cassgtz: I may still need this for the next task, can I just comment it out? @aaronshiel

client/src/pages/feedback.tsx
aaronshiel: yeah that sounds good!

cypress/cypress/integration/record-queue.spec.ts
aaronshiel: yeah np

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+  open: boolean;
+  mentor: Mentor;
+  userQuestion?: string;
+  accessToken: string;
+}): JSX.Element {
+  const { handleClose, open, userQuestion, mentor, accessToken } = props;
+  const classes = useStyles();
+  const [selectedSubject, setSelectedSubject] = useState<Subject>();
+  const [selectedCategory, setSelectedCategory] = useState<Category>();
+  const [selectedTopic, setSelectedTopic] = useState<Topic>();
```
aaronshiel, line 79: Questions can typically be part of multiple topics, I'd ask Ben if you should handle mapping the question to multiple topics or just 1 in this modal

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+      _id: uuid(),
+      question: customQuestion,
+      type: QuestionType.QUESTION,
+      name: UtteranceName.NONE,
+      clientId: uuid(),
+      paraphrases: [],
+      mentor: mentor._id,
+    };
+    // create newSubjectQuestion : to add to db
+    const emptyTopic: Topic = { id: "", name: "", description: "" };
```
aaronshiel, line 96: This won't work and shouldn't be done, this will throw errors since empty id's are not allowed and we shouldn't be creating "empty" documents

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+    const newSubjectQuestion: SubjectQuestionGQL = {
+      question: newQuestion,
+      topics: selectedTopic != undefined ? [selectedTopic] : [emptyTopic],
+      category: selectedCategory,
+    };
+    // add to DB
+    addOrUpdateSubjectQuestions(
+      selectedSubject?._id || "",
```
aaronshiel, line 104: Instead of accessing a state variable, I'd pass this in as a required parameter, so if the variable doesn't exist, then this function (okButtonClicked) shouldn't be called at all

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+      [newSubjectQuestion],
+      accessToken
+    );
+    // add to record queue
+    addQuestionToRecordQueue(newQuestion._id, accessToken);
+    // close modal
+    if (handleClose != undefined) {
```
aaronshiel, line 111: handleClose shouldn't ever be undefined (see other review comment above related to this function)

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+                          event: React.ChangeEvent<{ value: unknown }>
+                        ) => {
+                          setSelectedSubject(
+                            mentor.subjects.find((a) => {
+                              return a._id === event.target.value;
+                            })
+                          );
+                        }}
+                      >
+                        <option value={undefined} selected disabled>
```
aaronshiel, line 170: do you need to explicitly specify that value={undefined}? You can probably just remove that (I see that's what you did below), you usually want to avoid setting this explicitly to undefined

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+                          </option>
+                        ))}
+                      </Select>
+                    </FormControl>
+                    <FormControl>
+                      <InputLabel>Category</InputLabel>
+                      <Select
+                        // a dropdown for the category
+                        data-cy="category-drop-down"
+                        value={selectedSubject?.categories} // all the subject's categories
```
aaronshiel, line 189: It's probably better to just not render the category and topic dropdowns until a subject is selected. What's the behavior of this dropdown if a subject isn't selected yet?

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+                  onClick={handleClose}
+                  data-cy="modal-close-btn"
+                  variant="contained"
+                  color="primary"
+                  component="span"
+                >
+                  Cancel
+                </Button>
+                <Button
+                  onClick={() => okButtonClicked(customQuestion || "")}
```
aaronshiel, line 251: customQuestion isn't ever undefined, so you can remove the || ""

client/src/pages/feedback.tsx
```diff
@@ -116,16 +132,89 @@ const columnHeaders: ColumnDef[] = [
...
+  } = props;
+  const [selectedAnswerStatus, setSelectedAnswerStatus] = useState<Status>(); // for disabling/enabling queue button
+  const [selectedAnswerID, setSelectedAnswerID] = useState<string>();
+
+  // function to add/remove from queue
+  async function queueButtonClicked(
+    selectedAnswerID: string,
+    accessToken: string
+  ) {
+    if (queueList?.includes(selectedAnswerID)) {
```
aaronshiel, line 198: queueList isn't an optional param, so there shouldn't be any need for the ?

client/src/pages/feedback.tsx
```diff
@@ -116,16 +132,89 @@ const columnHeaders: ColumnDef[] = [
...
+      setQueueList(
+        await removeQuestionFromRecordQueue(selectedAnswerID, accessToken)
+      );
+    } else if (selectedAnswerID == undefined || selectedAnswerID == "") {
+      setOpenModal(true);
```
aaronshiel, line 203: This should have a better name, maybe "setCustomQuestionModalOpen", if that's what it's doing, not sure

client/src/pages/feedback.tsx
```diff
@@ -202,6 +302,30 @@ function FeedbackItem(props: {
             <IconButton onClick={() => onUpdateAnswer(undefined)}>
               <CloseIcon />
             </IconButton>
+
+            {accessToken ? (
```
aaronshiel, line 306: why not just also check for selectedAnswerID here aswell isntead of doing `|| ""` below

client/src/pages/feedback.tsx
```diff
@@ -202,6 +302,30 @@ function FeedbackItem(props: {
...
+              <Button
+                data-cy="queue-btn"
+                color="primary"
+                disabled={selectedAnswerStatus == Status.COMPLETE}
+                onClick={() => {
+                  queueButtonClicked(selectedAnswerID || "", accessToken);
+                }}
+              >
+                {queueList?.includes(selectedAnswerID || "")
```
aaronshiel, line 315: no need for ? here since queueList should never be undefined

client/src/pages/feedback.tsx
```diff
@@ -232,10 +356,12 @@ function FeedbackPage(): JSX.Element {
     isLoading: isMentorLoading,
     error: mentorError,
   } = useActiveMentor();
-
+  const { state: loginState } = useWithLogin();
   const mentorId = getData((state) => state.data?._id);
+  const Mentor = getData((state) => state.data);
```
aaronshiel, line 361: Mentor --> mentor

variable shouldn't be uppercase

client/src/components/feedback/edit-question-for-queue-modal.tsx
cassgtz: A topic list is required to create a Subject Question, but topics are optional fields in the modal. How do we handle that? Either we make topics required or use a different method in adding the new question to the db?

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -255,7 +251,9 @@ function EditQuestionForQueueModal(props: {
                   component="span"
                   disabled={
                     selectedSubject == undefined ||
```
aaronshiel, line 259: this can all be shortened to 
`!selectedSubject || !selectedCategory || !customQuestion`

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+  mentor: Mentor;
+  userQuestion: string;
+  accessToken: string;
+}): JSX.Element {
+  const { handleClose, open, userQuestion, mentor, accessToken } = props;
+  const classes = useStyles();
+  const [selectedSubject, setSelectedSubject] = useState<Subject>();
+  const [selectedCategory, setSelectedCategory] = useState<Category>();
+  const [customQuestion, setCustomQuestion] = useState<string>(userQuestion);
+  const selectedTopics: Topic[] = [];
```
aaronshiel, line 80: I'd be wary of storing data that changes as a class variable. Changes to class variables do not trigger re-renders of the page. So, if anything relies on this variable to display anything, this has to be a state variable. Else, it should be fine.

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,278 @@
...
+      category: selectedCategory,
+    };
+    // add to DB
+    addOrUpdateSubjectQuestions(
+      selectedSubject._id,
+      [newSubjectQuestion],
+      accessToken
+    );
+    // add to record queue
+    addQuestionToRecordQueue(newQuestion._id, accessToken);
```
aaronshiel, line 106: When newQuestion is sent to GQL via addOrUpdateSubjectQuestions, its _id gets replaced with a new _id, and that new _id gets returned by addOrUpdateSubjectQuestions (along with all other subject question data). You should get the question _id from what is returned by addOrUpdateSubjectQuestions and add that to record queue.

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,278 @@
...
+                  {selectedSubject ? (
+                    <div
+                      className={classes.inputField}
+                      style={{ textAlign: "left" }}
+                    >
+                      <Autocomplete
+                        // an autocomplete kwd-style area for "topics"
+                        data-cy="topic-selector"
+                        multiple
+                        options={selectedSubject?.topics || []}
```
aaronshiel, line 220: no ? needed here since selectedSubject is already checked for

client/src/components/feedback/edit-question-for-queue-modal.tsx
```diff
@@ -0,0 +1,278 @@
...
+                  <Button
+                    onClick={() =>
+                      okButtonClicked(customQuestion, selectedSubject)
+                    }
+                    data-cy="modal-OK-btn"
+                    variant="contained"
+                    color="primary"
+                    component="span"
+                    disabled={
+                      !selectedSubject || !selectedCategory || !customQuestion
```
aaronshiel, line 264: you can remove !selectedSubject from here all together since it's already checked

client/src/pages/feedback.tsx
```diff
@@ -212,6 +308,30 @@ function FeedbackItem(props: {
             <IconButton onClick={() => onUpdateAnswer(undefined)}>
               <CloseIcon />
             </IconButton>
+
+            {accessToken ? (
+              <Button
+                data-cy="queue-btn"
+                color="primary"
+                disabled={selectedAnswerStatus == Status.COMPLETE}
```
aaronshiel, line 316: ===

cypress/cypress/integration/record-queue.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    cy.get("[data-cy=category-drop-down]").click();
+    cy.get("[data-cy=Category-option-category1]").should("be.visible");
+    cy.get("[data-cy=Category-option-category3]").should("be.visible");
+    cy.get("[data-cy=Category-option-category1]").click();
+
+    cy.get("[data-cy=topic-selector]").click();
+    cy.get("[data-cy=Topic-option-back-topic1-id]").should("be.visible");
+    cy.get("[data-cy=Topic-option-back-topic2-id]").should("be.visible");
+  });
+  it.only("Modal creates new question", () => {
```
aaronshiel, line 59: remove .only

client/src/components/feedback/edit-question-for-queue-modal.tsx
cassgtz: addOrUpdateSubjectQuestions  returns AddOrUpdateQuestionGQL[], do you know if the new one always gets added to the end of that array? I can access the last element and get the ID that way

PR #208, opened 2022-06-15
MyMentor Splash Screen Overlay

Megan-J:


client/src/components/home/index.tsx
```diff
@@ -378,6 +385,11 @@ function HomePage(props: {
           reviewAnswerState.clearError();
         }}
       />
+      <NotificationDialog
+        title={"This page is for setting up your mentor!"}
+        open={!hasSeenSplash}
+        closeDialog={() => userSawSplashScreen("")}
```
aaronshiel, line 391: This needs to pass an accessToken to the userSawSplashScreen function instead of the empty string

client/src/store/slices/login/useWithLogin.ts
```diff
@@ -12,10 +12,12 @@ import * as loginActions from ".";
 interface UseWithLogin {
   state: loginActions.LoginState;
   login: (accessToken: string) => void;
+  userSawSplashScreen: (myMentorSplash: string) => void;
```
aaronshiel, line 15: the param here is an access token, the param name should reflect that

client/src/store/slices/login/useWithLogin.ts
```diff
@@ -52,6 +54,12 @@ export function useWithLogin(): UseWithLogin {
     }
   }
 
+  // Call this function when user clicks close on the dialog
+  function userSawSplashScreen(firstTimeTracking: string) {
```
aaronshiel, line 58: the param here is an access token, the param name should reflect that

aaronshiel (changes requested):

PR #211, opened 2022-06-21
Sentry logs redux api exceptions

aaronshiel:
Redux reducers are catching their respective api exceptions, so have to either re-throw or log manually in order to reach sentry, I chose to log manually


aaronshiel:
Just saw that I should be able to add sentry logging as middleware, need to change

aaronshiel:
https://github.com/mentorpal/mentor-admin/pull/212

PR #212, opened 2022-06-21
sentry added to redux middleware to log errors

aaronshiel:


client/src/store/store.ts
```diff
@@ -10,6 +10,16 @@ import loginReducer from "./slices/login";
 import configReducer from "./slices/config";
 import mentorReducer from "./slices/mentor";
 import questionsReducer from "./slices/questions";
+import * as Sentry from "@sentry/react";
+
+const sentryEnhancer = Sentry.createReduxEnhancer({
+  actionTransformer: (action) => {
+    if (action.error) {
+      Sentry.captureException(action.error.message);
```
mbsimonovic, line 18: think you're supposed to pass in an actual Error instance, so probably `action.error`

mbsimonovic (approved):

PR #222, opened 2022-07-07
Add config page for updating config settings

kycarr:


client/src/hooks/graphql/use-with-mentor-panel.tsx
```diff
@@ -0,0 +1,68 @@
...
+  editedData: MentorPanel | undefined;
+  isEdited: boolean;
+  isLoading: boolean;
+  isSaving: boolean;
+  error: LoadingError | undefined;
+  editData: (d: Partial<MentorPanel>) => void;
+  saveData: () => void;
+}
+
+export function useWithMentorPanel(
```
aaronshiel, line 22: I don't see this hook used anywhere, is it worth keeping?

aaronshiel (dismissed):

